PR #55, opened 2017-04-10
Changed codebase release URLs to be version_number and codebase identifier

cpritcha:
- Updated Codebase and CodebaseRelease serializers to match serialization of Jobs and Events
- Updated Codebase to have update and create routes
- Added Download, Follow and Share buttons to the Codebase retrieve page


django/library/urls.py
```diff
@@ -1,12 +1,21 @@
...
 
 router = SimpleRouter()
 router.register(r'codebases', CodebaseViewSet, base_name='codebase')
 router.register(r'codebases/(?P<identifier>\w+)/releases', CodebaseReleaseViewSet, base_name='release')
 
-urlpatterns = router.urls + [
-    url(r'^codebase/add', TemplateView.as_view(template_name='library/codebase/add.jinja'), name='codebase-add'),
+router.register(r'codebases', CodebaseViewSet)
+router.register(r'codebases/(?P<pk>[0-9]+)/releases', CodebaseReleaseViewSet)
+
```
alee: `urlpatterns = router.urls + create_edit_routes(prefix=Codebase._meta.object_name.lower() + 's', model=Codebase, lookup_field='pk', lookup_regex=r'\d+')`

django/library/views.py
```diff
@@ -12,29 +13,26 @@
 
 class CodebaseViewSet(viewsets.ModelViewSet):
     lookup_field = 'identifier'
-    multiple_lookup_fields = ('identifier', 'uuid', 'pk')
+    lookup_url_kwarg = 'pk'
```
alee: we should use the default `identifier` here to not break the expectation that the lookup kwarg matches the underlying model attribute

django/library/views.py
```diff
@@ -26,13 +25,14 @@ def template_name(self):
...
     queryset = CodebaseRelease.objects.all()
     serializer_class = CodebaseReleaseSerializer
+    pagination_class = SmallResultSetPagination
 
     def get_queryset(self):
         resolved = resolve(self.request.path)
-        codebase_id = resolved.kwargs['pk']
-        return self.queryset.filter(codebase_id=codebase_id)
\ No newline at end of file
+        identifier = resolved.kwargs['pk']
```
alee: lookup_kw_arg should be `identifier`

django/library/serializers.py
```diff
@@ -37,7 +37,8 @@ class CodebaseReleaseSerializer(serializers.ModelSerializer):
     programming_languages = home_serializers.TagSerializer(many=True)
 
     def get_relative_url(self, obj) -> str:
-        return reverse_lazy('library:codebaserelease-detail', kwargs={'release_pk': obj.codebase_id, 'pk': obj.id})
+        return reverse_lazy('library:codebaserelease-detail',
```
alee: Would recommend delegating to `CodebaseRelease.get_absolute_url` instead and define it there centrally. See https://docs.djangoproject.com/en/1.11/ref/models/instances/#get-absolute-url for more details

PR #107, opened 2017-08-14
Webpack rehaul

cpritcha:
I've rebuilt the webpack config to not use vue files and use webpack-blocks. It it much smaller now.

Bootstrap released a beta a days so I also upgraded bootstrap. Some CSS changes are still needed to make the site the same as it was before (the form classes and secondary button classes still need to changed).

I couldn't figure out how to get git to recognize my file renames and changes from `.vue` to `.ts` so there are more deletes and creates than necessary.

@alee


alee:
`git mv <old-path> <new-path>` usually works on the CLI.

PR #190, opened 2018-02-01
Query calendar using postgres timestamp ranges

cpritcha:


alee:
This is clever! But I don't find it to be an improvement over the original in terms of clarity / conciseness / simplicity. 

It looks like the changes to `core/views.py` should be included however..

PR #241, opened 2018-02-20
Improve share urls

cpritcha:
Note that this has some schema changes (set default field value of share_url to None and added a uniqueness constraint on share_url) but they should not impact the existing website.

### DIFF
--- a/django/library/fs.py
+++ b/django/library/fs.py
@@ -355,6 +355,10 @@ class CodebaseReleaseFsApi:
         return self.rootdir.joinpath('archive.zip')
 
     @property
+    def review_archivepath(self):
+        return self.rootdir.joinpath('review_archive.zip')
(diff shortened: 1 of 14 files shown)


django/library/fs.py
```diff
@@ -448,8 +452,12 @@ def initialize_codemeta(self, path=None, metadata=None):
             json.dump(metadata, codemeta_out)
         return True
 
+    def retrieve_review_archive(self):
+        shutil.make_archive(str(self.rootdir.joinpath('review_archive')), format='zip', root_dir=str(self.sip_contents_dir))
```
alee: Not a huge fan of having multiple ways to generate the review_archivepath components with and without the .zip extension but that's the price of `shutil.make_archive` I guess..

Issue #437, opened 2019-04-10
peer review validation


alee:
peer review requests shouldn't go through if there is no code / documentation to review

alee:
@MarcoAJanssen suggests that we also run full publish validation on the model to ensure it has all the metadata fields required for publishing as well

alee:
This is almost working, but not quite. When I try to request a peer review for a model that isn't publish()-able, it returns without requesting the peer review, but I don't see any of the appropriate / relevant error messages. Instead, I only see the relevant validation error messages if I request peer review again, e.g.:

![peer-review-errors](https://user-images.githubusercontent.com/22534/65663307-46d23280-dfeb-11e9-931e-54c015efe5f0.PNG)

PR #472, opened 2019-08-13
added conditional field requirements for peer reviewer validation

chrstngyn:
if checkbox is checked, comments field is enabled


chrstngyn:
TODO:

- [ ] verify metadata on peer reviewer form submission

alee:
Thanks @chrstngyn ! Just to clarify, in #437 we'd like to validate that the model has the necessary metadata when the model author *submits* a review request, otherwise the review request doesn't go through and we display an appropriate error message to the model author. So the codebase should have an OSS license, uploaded documentation, uploaded code, contributors properly added, etc. In other words it should meet the same metadata requirements as it takes to publish the model.

Sorry for the confusion, I should've made #437 more clear..

PR #485, opened 2019-09-16
Codebase replication field

### LINKED ISSUE #179: replication checkbox state should toggle replication_text field visibility
Checking the replication checkbox in /codebases/add/ should make a replication_url field magically appear in /codebases/add

chrstngyn:
while user is archiving their model, if replication checkbox is checked, a text field for submitting url / doi / citation to the original model becomes visible. user must submit url / doi / citation if their model is a replication.

resolves #179

### DIFF
--- a/frontend/src/pages/codebase/Edit.vue
+++ b/frontend/src/pages/codebase/Edit.vue
@@ -10,8 +10,10 @@
                     :required="config.is_replication"
                     help="Is this model a replication of a prior computational model?">
         </c-checkbox>
-        <c-input v-if="is_replication" v-model="replication_text" :errorMsgs="errors.replication_text" name="replication text" label="URL / DOI / Citation"
-        help="URL / DOI / citation for the original model being replicated">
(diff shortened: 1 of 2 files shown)


frontend/src/pages/codebase/Edit.vue
```diff
@@ -10,6 +10,11 @@
                     :required="config.is_replication"
                     help="Is this model a replication of a prior computational model?">
         </c-checkbox>
+        <c-input v-if="is_replication" v-model="replication_text" :errorMsgs="errors.replication_text"
+                 name="replication_url" label="URL / DOI / Citation"
+                 help="URL / DOI / citation for the original model being replicated"
+                :required="config.replication_text">
```
alee: Should this be `:required="config.is_replication"` instead so the text is required only if `is_replication` is checked? 

Also, the `:required` for the `is_replication` checkbox above this component seems odd to me, it's only required if it is checked?

PR #498, opened 2019-12-05
Only allow file codebase image uploads to be gif, jpeg or png

cpritcha:
Previously any image/* types were allowed.

Testing:

- tested client side filters in Safari, Chrome, Firefox. Filters don't work in Firefox
- server side now rejects any file type without gif, jpeg, png headers
- tests codebase file uploads to make sure they still work

### DIFF
--- a/django/core/fs.py
+++ b/django/core/fs.py
@@ -36,7 +36,8 @@ def is_archive(path: str):
 
 def is_image(path: str):
     try:
-        return imghdr.what(path)
+        filetype = imghdr.what(path)
+        return filetype in ['jpeg', 'png', 'gif']
     except:
         return None
 
--- a/frontend/src/components/upload.ts
+++ b/frontend/src/components/upload.ts
@@ -29,7 +29,8 @@ type UploadInfo = UploadSuccess | UploadProgress | UploadFailure;
         <div class="d-flex justify-content-between">
             <div>
                 <label :for="uploadId"><div class="btn btn-primary">Upload a file</div></label>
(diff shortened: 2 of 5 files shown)


django/core/fs.py
```diff
@@ -36,7 +36,8 @@ def is_archive(path: str):
 
 def is_image(path: str):
     try:
-        return imghdr.what(path)
+        filetype = imghdr.what(path)
+        return filetype in ['jpeg', 'png', 'gif']
```
alee: let's define the accepted image types in one place as a constant (maybe in settings?)

`ACCEPTED_IMAGE_TYPES = ['jpeg', 'png', 'gif']`

we can use it here and in the exception format string that you are building up `library/models.py`

frontend/src/components/upload.ts
```diff
@@ -29,7 +29,8 @@ type UploadInfo = UploadSuccess | UploadProgress | UploadFailure;
         <div class="d-flex justify-content-between">
             <div>
                 <label :for="uploadId"><div class="btn btn-primary">Upload a file</div></label>
-                <input class="invisible" :id="uploadId" type="file" @change="handleFiles($event)">
+                <input class="invisible" :id="uploadId" type="file"
+                       @change="handleFiles($event)" :accept="acceptedFileTypes">
```
alee: was acceptedFileTypes missing earlier?

cpritcha: Yes. It doesn't properly filter on Firefox (all files included) but works on Safari and Chrome

PR #593, opened 2022-06-08
Made formatting changes to the education pages based on some of the feedback we have received.

seanbergin:
Made formatting changes based on some of the feedback we have received.
- changed all of the 'good' practices to 'responsible' practices for consistency


alee (changes requested):
Thanks Sean! I'll need to do a little more work here because this management command should be idempotent (i.e., running the command results in the same outcome in the database and on the site, but changing the URL from `good-practices` to `responsible-practices` means we need to add some guardrails to look up the old slug if it exists and remove or update it instead).

PR #595, opened 2022-06-10
Request download feature

chrstngyn:

PR #598, opened 2022-07-11
feat: add summer digest to comses net

chrstngyn:
- add pdf to static/digest/
- create list item for summer digest in digest.jinja template


alee (changes requested):
Thanks @chrstngyn ! The digest related changes look perfect but it looks like changes from #595 got included as well though? We should probably keep those isolated to the branch

alee (approved):
Perfect!

PR #617, opened 2023-02-10
refactor build process

alee:
switch to Makefile based build / deployment for consistency across port of mars etc.


alee:
For some reason the js container's webpack-stats.json isn't visible to Python app server's container. Could be the added `--chown=node:node`?

PR #620, opened 2023-02-27
fix social connection issues

sgfost:
- Profile edit form redirects to /accounts/social/connections as a central form (which now correctly uses POST) for managing account connections which looks like:
Screenshot from 2023-02-27 12-53-12
alternatively, could render the direct links in the template before mounting the vue component form

- errors are now display properly in the connections page instead of rendering a method
- links to password reset page when trying to remove a lone connection from an account with no password set

### DIFF
--- a/django/core/jinja2/socialaccount/connections.jinja
+++ b/django/core/jinja2/socialaccount/connections.jinja
@@ -19,7 +19,7 @@
         {{ csrf_input }}
 
         {% if form.non_field_errors() %}
-            <div id="errorMsg">{{ form.non_field_errors }}</div>
+            <div id="errorMsg">{{ form.non_field_errors() }}</div>
         {% endif %}
         <div class='form-group row'>
             <div class='col-10'>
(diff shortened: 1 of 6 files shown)


frontend/src/components/profile/Edit.vue
```diff
@@ -35,22 +35,30 @@
...
+              </a>
+              <a title="Edit connection" :href="socialConnectionsURL">
+                <i class="float-right fas fa-edit"></i>
+              </a>
             </span>
             <span v-else>
-              <a title="orcid" href="/accounts/orcid/login/?process=connect"
-                ><span class="text-gray fab fa-orcid"></span> Connect your ORCID account</a
-              >
+              <a title="github" :href="socialConnectionsURL">
```
alee, line 46: We could either get rid of the  `a title` here and elsewhere in this text or add more semantic info to them like "sign in with orcid", "sign in with github" etc

django/core/jinja2/socialaccount/connections.jinja
```diff
@@ -37,6 +37,13 @@
             </div>
         </div>
         <button type="submit" class='btn btn-sm btn-primary'><span class='fa fa-trash'></span> Remove</button>
+        {% if form.accounts|length == 1 and not request.user.has_usable_password() %}
```
alee, line 40: good catch!

django/core/jinja2/socialaccount/connections.jinja
```diff
@@ -19,7 +19,7 @@
         {{ csrf_input }}
 
         {% if form.non_field_errors() %}
-            <div id="errorMsg">{{ form.non_field_errors }}</div>
+            <div id="errorMsg">{{ form.non_field_errors() }}</div>
```
alee, line 22: :100:

alee (commented):
LGTM, could fix the <a href title=...> attributes or leave for another day

frontend/src/components/profile/Edit.vue
sgfost: cleaned up the markup around here + some other consistency changes

alee (approved):
Tested on staging - LGTM, thanks for this @sgfost !

PR #625, opened 2023-03-13
vue 3 initial setup

sgfost:
Setting up `frontend-next` directory/container to be used as a base for porting over our vue 2 widgets to vue 3. Set up in a similar way to the current frontend with improvements where possible

uses:
- vue 3.2
- vite
- django-vite (https://github.com/MrBin99/django-vite)
- bootstrap 5 (https://getbootstrap.com/docs/5.0/migration)

comses/planning#95


sgfost:
This is just about ready as a base, just need to do some testing for running vite/django-vite in production mode

sgfost:
Collecting built assets for production seems to be in order, at least with what I can test locally.

CI workflows now feel a bit messy, we probably don't need to care about running checks on the 'legacy' frontend code anymore. And we could either integrate frontend checking/testing into the main [Docker CI](https://github.com/comses/comses.net/actions/workflows/docker-build.yml) workflow or keep a separate frontend CI and let the main one run server tests/any integration tests which would ideally at a minimum check to make sure things get 'mounted' correctly

sgfost:
We may be better off keeping this as a separate branch and then folding in once the 'legacy' code has been replaced. Basically a trade off between dealing with a _potentially_ error prone setup for any critical releases or a large(r) merge later on. Sounds easier to kick the production testing can down the road, so I'd lean towards keeping it out of main until its done. Any thoughts @alee?

PR #627, opened 2023-03-17
Modified regexes to account for additional cases

hwelsters:
- Corrected misclassification of R
wrong-R

- Modified Python regex to account for IronPython

### DIFF
--- a/django/curator/models.py
+++ b/django/curator/models.py
@@ -149,7 +149,7 @@ PLATFORM_AND_LANGUAGE_MATCHERS = [
     Matcher("Objective-C", pl_regex(r"objective(:?[-\s]+)?c")),
     Matcher("Pandora", pl_regex("pandora")),
     Matcher("Powersim Studio", pl_regex("powersim\s+studio")),
-    Matcher("Python", pl_regex("python")),
+    Matcher("Python", pl_regex(r"ironpython|python")),
     Matcher("R", pl_regex("r")),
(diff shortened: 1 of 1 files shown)


django/curator/models.py
```diff
@@ -149,8 +149,8 @@ def pl_regex(name, flags=re.I):
     Matcher("Objective-C", pl_regex(r"objective(:?[-\s]+)?c")),
     Matcher("Pandora", pl_regex("pandora")),
     Matcher("Powersim Studio", pl_regex("powersim\s+studio")),
-    Matcher("Python", pl_regex("python")),
-    Matcher("R", pl_regex("r")),
+    Matcher("Python", pl_regex(r"ironpython|python")),
```
alee, line 152: Should the regex instead allow for arbitrary \*python\* to map to `Python`? pl_regex doesn't allow arbitrary characters in front for some reason - I'm sure it was by design, but not sure why.

IronPython is a pretty esoteric distribution, not even listed there.

https://wiki.python.org/moin/PythonDistributions

R regex looks fine with me to prevent the `R2016a` from being caught, but we should double check this against the data in tags.csv (or all codebase release programming language tags)

alee (changes requested):
Thanks for the PR! 

I would prefer generalizable improvements instead of one-off fixes here so will defer on merging this until we have more information on what other corner cases we might want / need to deal with

PR #630, opened 2023-03-20
Add digest archives

sgfost:
Adds digest archives up to Summer 2022 as static files (pdf)

Reworked `DigestView` to be a simpler list view of `DigestArchive`s which are basically just metadata and a pointer to the static file.

`DigestArchive`s are automatically created via a new management command, given the name of the file added to the static directory is formatted well-enough

resolves comses/planning#96


django/home/models.py
```diff
@@ -940,6 +940,32 @@ def __str__(self):
         return "submission {} by {}".format(self.title, self.submitter)
 
 
+class DigestArchive(index.Indexed, models.Model):
```
alee, line 943: Might be good to also include a OneToMany of Contributors to keep track of the people who worked on the Digest.

django/home/models.py
```diff
@@ -940,6 +940,32 @@ def __str__(self):
...
+    represents a single issue of the quarterly digest with a pdf archive in the static directory
+    """
+
+    class Seasons(models.IntegerChoices):
+        SPRING = 1, _("Spring")
+        SUMMER = 2, _("Summer")
+        FALL = 3, _("Fall")
+        WINTER = 4, _("Winter")
+
+    year_published = models.IntegerField()
```
alee, line 954: should we also keep track of a `date_created` or something equivalent for internal book-keeping? Or we could have a single `publication_date` and extract the year from it

django/home/models.py
alee, line 943: How about naming this `ComsesDigest` instead? An Archive is typically a collection of things, but an instance of this class represents a single issue of the comses digest.

alee (changes requested):

django/home/models.py
alee: also, a doi or doi_url might be a good future-proof field in the event that we ever start DOI-ing each comses digest

alee: as well as title / summary fields so we can have a slightly better display in the comses.net/digest list view

django/home/models.py
alee: to clarify, something like `publication_date = models.DateField()`

not sure if it's necessary to keep `last_modified` + `date_created` auto fields here like we do for other comses models where edit date provenance is important, thoughts?

django/home/models.py
alee: I guess technically it could be an ArchivedComsesDigest but overall I think ComsesDigest might work better. Happy to entertain other ideas though!

django/home/models.py
sgfost: Originally there was a `publication_date` field but it seemed to me the publication date may be ambiguous. On second thought, I suppose the end date of the given period may be it, or close enough? 

Screenshot from 2023-03-23 09-44-23

django/home/models.py
alee: Sorry for the spam! I should stop late night reviewing. Let's nix the title / summary idea, seems unnecessary on second thought

django/home/models.py
alee: Yep, the end date of the period the digest covers. Now it's always 12/16-3/15, 3/16-6/15, 6/16-9/15, 9/16-12/15 though in the early days we'd make the interval non-standard to cover whenever we actually got around to sending the thing out

django/home/models.py
alee: As for Contributors: mostly (all?) John Murphy for the early Digests. In the more recent ones we've had a rotating guest editorship across the comses executive board but tends to default to Sean Bergin if we can't find anyone. Manuela and Kelly also do a lot of work to put it together, we should definitely include them as well as Laura Swantek for some of the previous digests. We don't have to go all out on this and can fill in the gaps later too.

Was trying to think up metadata that could spruce up the list of Digests (publish date, list of contributors, etc)

sgfost: Sounds good, contributors is just a blank m2m now. Editor seems to be named on each one so I suppose we could start with that, just need to think of the best way to input this now and in the future, I suppose an index file would work

alee (approved):
LGTM, thanks Scott! Added a few minor changes.

sgfost:
Sweet, thanks for the style changes

Issue #624, opened 2023-03-08
improve "list entry" components ux


sgfost:
- affiliations entry for codebase contributors should use the ROR org search component
- `EditTextList (textitem.ts)` and `EditOrgList (orgitem.ts)` components should more clearly indicate that you need to press enter to add an item (similar to how multiselect does), ideally within the input box. These could also automatically add the contents of the input to the list when focus from the input is lost (clicked/tabbed off, etc)

Issue #489, opened 2019-09-25
frontend: refactor createFormValidator


alee:
right now the `createFormValidator` works but is a bit of a convoluted mess that is possibly only understood by @cpritcha. We should see if there are better ways of abstracting its responsibilities:

1. form validation
2. others?

alee:
as a first step we probably need to just document it better so that we can figure out how to use it properly. I think we figured this out as needing to modify the yup() schema

Issue #451, opened 2019-05-13
refactor handler.ts Handlers


alee:
Initial draft design of interacting components:

- API knows the endpoint and how to invoke it and with what parameters and drives the process
- ServerResponseHandler builds coherent data structures out of a server response for use by the Component
- Components are responsible for stitching together the ServerResponseHandler and doing stuff based on its state (dismiss a modal, redirect on success, display a message somewhere on the DOM, display validation errors)

Issue #319, opened 2018-04-17
Frontend Forms


cpritcha:
vuejs has a number of form builders now

- https://github.com/jarvelov/vue-form-json-schema
- https://github.com/vue-generators/vue-form-generator
- https://github.com/formschema/native

It would be good to use a form builder for our forms instead of the manual process we're using now (which has some large templates and repeated logic). We would need to make sure that the form builder supports the same validation features that we're using right now.

PR #639, opened 2023-04-06
Vue3 form components

sgfost:
Adds:
- Form management API and requests API
- Form component library (inputs)
- Select forms implemented as usage examples


frontend-vue3/src/components/form/FormDatePicker.vue
```diff
@@ -32,17 +33,18 @@ import FormLabel from "@/components/form/FormLabel.vue";
...
 
-export type SelectOptions = { value: any; label: string }[];
-
 export interface DatePickerProps {
   name: string;
   label?: string;
   help?: string;
   placeholder?: string;
   required?: boolean;
+  string?: boolean; // whether to use string or date for the model value
```
alee, line 42: for booleans I lowkey prefer prefixing with is/can to make it a predicate verb so `isRequired` `isString` etc. thoughts?

sgfost: Yeah that's an improvement. May be a better way to express these in particular since the names don't indicate all that well what they do

sgfost: changed `required` -> `indicateRequired` since it just draws an asterisk, removed `string` in favor of only dealing with `Date` in the datepicker

sgfost:
All form components (minus the placeholder markdown component) are finished, likely along with any major changes to the form api.

I'll put in place some component tests to finish off this PR and then work on some documentation for the different apis and then putting together the rest of the pages should be (relatively) straightforward

sgfost:
resolves comses/planning#106, resolves #319, resolves #489, resolves #451, resolves #251, resolves #624

frontend-vue3/src/composables/api/axios.ts
```diff
@@ -0,0 +1,210 @@
...
+  return value && typeof value === "string" && ISODateFormat.test(value);
+}
+
+function convertDates(body: any) {
+  if (body === null || body === undefined || typeof body !== "object") {
+    return body;
+  }
+  for (const key of Object.keys(body)) {
+    const value = body[key];
+    if (isISODateString(value)) {
```
alee, line 204: could also use `Date.parse`, `try/catch` or are we only allowing iso date strings here?

frontend-vue3/src/composables/api/profile.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  async function update(id: string | number, data: any, options?: RequestOptions) {
+    return put(detailUrl(id), data, options);
+  }
+
+  async function uploadProfilePicture(id: string | number, file: any) {
+    const formData = new FormData();
+    formData.append("file", file);
+    return postForm(`${detailUrl(id)}upload_picture/`, formData, {
```
alee, line 34: maybe useAxios can also expose url building functions that hang off things like `detailUrl` - something like `buildUrl(detailUrl(id), 'upload_picture')` or an optional second argument path for a signature like `detailUrl(id, 'upload_picture')`

frontend-vue3/src/composables/api/revieweditor.ts
```diff
@@ -0,0 +1,47 @@
...
+   * Composable function for making requests to the reviews API
+   *
+   * @returns - An object containing reactive state of the request and helper functions for API requests
+   */
+
+  const baseUrl = "/reviews/";
+  const { state, get, post, put, detailUrl } = useAxios(baseUrl);
+
+  async function listInvitations(review_uuid: string) {
+    return get(`${detailUrl(review_uuid)}editor/invitations/`);
```
alee, line 15: Adding url building or concatenation to `useAxios` might make these methods cleaner as well

alee (commented):
This is a huge pile of excellent work, thanks Scott! I'm good to merge it into the `vue3` branch and continue moving forward there in a separate PR / commit stream but would leave it up to you. Left a few minor notes here and there. I think the main thing might be to rename the form components.

frontend-vue3/src/components/form/FormAlert.vue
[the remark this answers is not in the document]
sgfost: Sort of between a rock and a hard place w/ smurf naming and vue style guide/HTML collisions (https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names)

How about getting more descriptive and prepending/appending more specific names?:
- `Form<...>.vue` for non-field form-level stuff (really only alert for now) - `FormAlert.vue`
- `Field<...>.vue` for field-level 'helpers' like label/error - `FieldError.vue`
- `<...>Field.vue` for field components - `CheckboxField.vue`

Otherwise we're looking at more general prefixes, the vue 2 codebases uses `C` presumably for comses or custom. Here, we'd have to prefix the file name or add another script tag to set the name since `<script setup>` exports and imports the component as the file name automatically

frontend-vue3/src/components/form/FormOrgItems.vue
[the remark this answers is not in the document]
sgfost: Yep this is the one that uses the base lookup select/autocomplete + regular text inputs for a list entry. I think I'm going to use `List` instead of `Items` for this as well as the similar text items component

frontend-vue3/src/components/form/FormAlert.vue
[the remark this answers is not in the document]
alee: Ahh, gotcha - thanks for the style guide link! We could try something like `ValidationErrorAlert`. So TextArea and TextInput and things would conflict with native `<textarea>` tags or would it be translated to `text-area` which might also be confusing :sweat_smile:

frontend-vue3/src/composables/api/axios.ts
sgfost: Yeah the intent was to strictly parse ISO 8601 date strings and assume everything else is meant to be a string, the spec for `Date.parse` seemed far too loose since most implementations would parse a string like `"10 10 10"`

frontend-vue3/src/components/form/FormAlert.vue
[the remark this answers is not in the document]
sgfost: Yeah since you can use either `<text-area>` or `<TextArea>` or any case variation I think it would still cause a conflict there. I tried out my first idea for now: https://github.com/comses/comses.net/tree/db8c099f6a35f0d2b2b9e99a4404b0b5d817949c/frontend-vue3/src/components/form, lmk if you have any thoughts

frontend-vue3/src/components/form/FormOrgItems.vue
```diff
@@ -0,0 +1,189 @@
...
+import FormError from "@/components/form/FormError.vue";
+import FormPlaceholder from "@/components/form/FormPlaceholder.vue";
+import type { Organization } from "@/composables/api/ror";
+
+export interface TextItemsProps {
+  name: string;
+  label?: string;
+  help?: string;
+  placeholder?: string;
+  indicateRequired?: boolean;
```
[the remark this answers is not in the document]
sgfost: Looks like the fix that allows doing this just got added a few weeks ago (https://github.com/vuejs/core/pull/8083) so I'll note it down for when the next release comes

sgfost:
Super helpful review @alee, thanks. Everything here should be addressed. I'll merge this into `vue3` and then keep going straight on there for the most part

PR #643, opened 2023-05-01
option for metrics page refactor

sgfost:
- construct the various charts beforehand and pass in a map prop which is then used to display based on a radio button selection
- added 2 different charts: cumulative line + column for growing data and percentage area for comparison data
- added missing data table
- move styling to bootstrap

potential TODOs:
- [x] fix incoming server data (fill in date ranges)
- [x] add some form of indication for current year
- [ ] figure out why labels aren't showing up inside the area chart

### DIFF
--- a/frontend/src/pages/metrics/index.ts
+++ b/frontend/src/pages/metrics/index.ts
@@ -14,22 +14,22 @@ export interface Metric {
   series: TimeSeries[];
 }
 
-export interface MetricsData {
-  start_year: Metric;
-  total_members: Metric;
-  full_members: Metric;
-  total_codebases: Metric;
-  codebases_by_os: Metric;
-  codebases_by_platform: Metric;
-  codebases_by_language: Metric;
-  reviewed_codebases: Metric;
-  total_downloads: Metric;
-}
+export type MetricsData = Record<
+  | "start_year"
+  | "total_members"
+  | "full_members"
+  | "total_codebases"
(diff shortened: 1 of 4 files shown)


frontend/src/pages/metrics/index.ts
```diff
@@ -1,59 +1,37 @@
...
+  name: string;
+  data: number[];
+  type?: string;
+}
 
-  console.log("metricsData");
-  console.log(JSON.stringify(metricsData, null, 4));
+export interface Metric {
+  title: string;
+  y_label: string;
```
alee, line 12: should we consider using something like https://humps.readthedocs.io/en/latest/ to camelCase our snake cased python data?

alee (approved):
LGTM, is there any pending work here besides the labels? Otherwise I'm good to merge it in now.

The snake case to camel case has been bugging me a bit as it "infects" our JS codebase with inconsistent variable name casing but it's a minor pet peeve. I guess we could also snake case the JSON that we emit from our endpoints

frontend/src/pages/metrics/index.ts
sgfost: Should be all good, I'm guessing highcharts is deciding to not use labels due to some series being too small, however I haven't managed to find a way around this. 

Agreed on the casing though, its something I'd like to resolve in the v3 codebase, I like the idea of doing the conversion server-side though

PR #652, opened 2023-06-01
Frontend Vue 3 Upgrade

sgfost:
Adds a new frontend codebase written with [Vue 3 using the composition API](https://vuejs.org/guide/extras/composition-api-faq.html) and served/bundled with [Vite](https://vitejs.dev/). Additionally, [django-webpack-loader](https://github.com/django-webpack/django-webpack-loader) is replaced with [django-vite](https://github.com/MrBin99/django-vite) for the integration of bundled vue components into Django/jinja templates.

Components and APIs are either ported or rewritten and based off of the previous Vue 2 codebase. Unfortunately an incremental migration was not feasible so here we are with this massive merge request.

More detailed information can be found in the (currently expanding) [documentation for the codebase](https://github.com/comses/comses.net/wiki/Frontend-Components-(vue3)).

### Hot reloading/bundling

`Vite` (which uses [rollup](https://rollupjs.org/) for bundling) is used instead of `vue-cli` and `webpack`. This fixes/provides much faster hot-reloading in development mode and functions largely the same when bundling for production. `django-webpack-loader` is replaced by `django-vite` for integrating into the django backend, which functions similarly.

### Composition

The Vue composition API replaces the class-based options API style of the previous codebase. This extends to replacing class-based APIs with [composable functions](https://vuejs.org/guide/reusability/composables.html).

#### Form Management

Form management/validation is partially handled with [vorms](https://vorms.mini-ghost.dev/) which provides a very lightweight API that is very similar to [React Hook Form](https://react-hook-form.com/). [Yup](https://github.com/jquense/yup) is still used for validation. The idea is to provide a lightweight/reusable API for building forms that replaces the class/inheritance-based system in the previous codebase (which is no longer an option with Vue 3).

### UX Upgrades

Porting everything provided a great opportunity for design upgrades/refreshes where needed. Most notable changes include:
- Slightly rework search/filtering (codebases, profiles, jobs, events)
- Upgrades for complex form input components (particularly list entries)
- Slightly improved validation/error messaging for simple forms and add unsaved alerts
- Overhaul of release editor page to cut down on clutter and streamline the process (particularly editing contributors)

### Maintainability

Starting from scratch allowed some significant re-organization and reduction of the amount of code/dependencies. The hope is that the architecture and individual components should be easier to understand/work on.

##  Remaining Related Work

- [ ] more thoroughly test in a prod. environment (staging)
[...]

PR #653, opened 2023-06-01
Wagtail 5 search fixes

sgfost:
resolves comses/planning#123

- [x] replace/migrate `Query` model: https://docs.wagtail.org/en/latest/releases/5.0.html#wagtailsearch-query-has-moved-to-wagtail-contrib-search-promotions

**management/migrations**:
- move wagtail search query models to new location
  - `./manage.py migrate wagtailsearchpromotions`
  - `./manage.py copy_daily_hits_from_wagtailsearch`
    - **(might want to archive before copying over due to the huge amt of records)**
- archive daily query hits
  - [search/0001](https://github.com/comses/comses.net/blob/0a2c50c4541547249cadd6bd1902b265188a296b/django/search/migrations/0001_initial_archived_queries.py) add `ArchivedQueryHits` (search hits grouped by year)
  - `./manage.py archive_search_query_hits` [[options](https://github.com/comses/comses.net/blob/0a2c50c4541547249cadd6bd1902b265188a296b/django/search/management/commands/archive_search_query_hits.py)]


sgfost:
9ab1528 breaking `check_list()` tests, but only for profile/event/job and not codebase..

sgfost:
Needed to use `search_backend.autocomplete()` only when we have a query that isnt `MatchAll`. Search functionality seems to be back in order now

sgfost:
This involves a few migration steps to move everything over to the new Query model location (https://docs.wagtail.org/en/latest/releases/5.0.html#wagtailsearch-query-has-moved-to-wagtail-contrib-search-promotions):

```
./manage.py migrate wagtailsearchpromotions
```
```
./manage.py copy_daily_hits_from_wagtailsearch
```
```
./manage.py searchpromotions_garbage_collect
```

alee:
should we merge this in before the vue3-replace PR #652 ?

sgfost:
Sure, that order makes a bit more sense

alee (commented):
After reading through https://docs.wagtail.org/en/stable/topics/search/searching.html more thoroughly I’m wondering if some of these autocomplete fields should be basic search fields instead..?

sgfost:
I think you're right, my fault assuming most of this was a drop-in replacement for `partial_match`.

On second thought, maybe the frontend merge can go first while this gets revised and something for archiving query hits gets added (in the works now)

sgfost:
Added a new command `archive_search_query_hits` which moves daily hits into a more condensed (grouped by year) table and then does the garbage collection on the daily hits and optionally queries table. This isn't scheduled to run currently but it could be a weekly/monthly thing

sgfost:
As far as I can tell, removing `partial_match` and using regular `SearchField`s with the current search views seems to work just fine. Maybe that is the way to go and we can reserve AutocompleteField for its intended use?

alee:
Sounds good to me! There's also support for facets now, so once we get tag cleaning done we can actually do some proper faceted searches across the model library

sgfost:
I did a bit of messing around with the wagtail facets and they are pretty easy to drop in and use, only tricky bit will be adding specific language/framework filters

PR #655, opened 2023-06-28
Event/job deleting and indicating expiration

### LINKED ISSUE #420: event + job delete buttons do not work
seems we never actually implemented this actually. need to send a DELETE and verify proper authentication

### LINKED ISSUE #582: visually demarcate expired events / jobs
instead of hiding them

sgfost:
- Implements deleting jobs and events by marking with an `is_deleted` field and excluding these from the queryset. A confirmation modal is shown from the existing delete button

- Adds an expiration indicator to jobs and events and now shows all results by default. What constitutes 'expired' (displayed as "past events") for events changed to an end date after today or start date after 7 days ago if end date is null
---

**migrations**:
- [core/0017](https://github.com/comses/comses.net/blob/ac8c3d78cdeb06c95ddb4175ea55a3cde1338d36/django/core/migrations/0017_job_event_is_deleted.py) add is_deleted to job/event
- [core/0018](https://github.com/comses/comses.net/blob/ac8c3d78cdeb06c95ddb4175ea55a3cde1338d36/django/core/migrations/0018_remove_event_date_times.py) change event `DateTimeField`s to `DateField`s

---

resolves #420
resolves #582


alee:
awesome, thanks @sgfost ! I'd like to have some test coverage here too, especially for the time bound and expiration logic. Might be a good side project for the student devs as well to become familiar with writing tests, since I would like for them to include tests on their NLP/ML work as well

sgfost:
agreed, I have some in the works since I wanted a bit more confidence in those Q objects

alee (changes requested):
Looks good, but a few minor notes:

We probably need to add some redirect guards, manually visiting jobs/<id>/delete/ takes you to an edit page for that job|event even if you don't have the right perms

https://staging.comses.net/jobs/592/delete/

for example

Would be good to clarify `timeuntil` usage, and maybe set it to `depth=1` for a more succinct string

sgfost:
Looks like that's the case for every edit page, never would have noticed. I'll put in something to stop that

django/core/jinja_config.py
```diff
@@ -182,6 +185,14 @@ def format_datetime(date_obj, format_string=FULL_DATE_FORMAT):
     return date_obj.strftime(format_string)
 
 
+def timeuntil(date):
+    if date is None:
+        return None
+    elif isinstance(date, str):
+        date = parse_datetime(date)
+    return timesince(date, timezone.now())
```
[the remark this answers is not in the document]
sgfost: Sure is, no clue what I was thinking here since this is just wrapping `timesince` for str input (and being used for time since)  and calling it `timeuntil`

Showing time until as well is a good idea, I'm thinking that could go where the expired badge is currently

sgfost:
Vying for the record on the highest force push to commit ratio on this one. I believe ac8c3d78cdeb06c95ddb4175ea55a3cde1338d36 fixes the issue with edit pages being accessible without perms, unless there is a reason that https://github.com/comses/comses.net/blob/ac8c3d78cdeb06c95ddb4175ea55a3cde1338d36/django/core/permissions.py#L21-L41 wasn't being used

django/library/serializers.py
sgfost, line 1: This looks like a frontend problem too, hitting save without anything filled out for a contributor besides a role sends bad a bad request. I'll add a fix for that too

alee: thanks! Sorry, just saw this and may have caused conflicts with some another set of commits to the UserSearch/ContributorSearch components to remove usernames from the display since they aren't that helpful for helping someone find an existing contributor in the system.

For some reason the institution / affiliation name is never displayed properly, probably something in the serializers or the autocomplete search?

alee (approved):
I think this is g2g after the frontend contributor patch is in

django/library/serializers.py
sgfost: I'll look into that as well, I'm working on some changes to `get_existing_contributor` atm so it does exact matching of firstname, lastname, and email instead

django/core/jinja_config.py
```diff
@@ -169,17 +175,44 @@ def add_field_css(field, css_classes: str):
...
         return None
-    d = parse_datetime(text)
-    return format_datetime(d, format_string)
+    datetime = parse_datetime(text)
+    return format_datetime(datetime, format_string)
 
 
-def format_datetime(date_obj, format_string=FULL_DATE_FORMAT):
-    if date_obj is None:
+def format_datetime(datetime, format_string=FULL_DATETIME_FORMAT):
```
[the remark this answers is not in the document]
sgfost: After I figure out the affiliations display on search I'll rebase on top of main to fix these. Lmk if you have any changes I should wait for

django/library/serializers.py
```diff
@@ -75,13 +75,12 @@ def get_existing_contributor(self, validated_data):
...
         else:
-            # filter by given_name and family_name if they were entered
-            name_filter = {
-                k: v
-                for k, v in validated_data
-                if k in ("family_name", "given_name") and v
+            # find an existing contributor that exactly matches the given name and email
+            contrib_filter = {
+                k: validated_data.get(k, "")
+                for k in ["given_name", "family_name", "email"]
```
alee, line 81: the dict comprehension was to create filter criteria like

 `{"email": "foo@example.com"}`

instead of 

 `{ "given_name": "", "family_name": "", "email": "foo@example.com" }`

wouldn't the latter prevent a record that matches the email but also has a given name and family name from matching?

sgfost: Oh right, we should always match emails, I was just thinking about name conflicts

maybe
```
if email:
  get by email
else:
  filter by name + blank email (?)
```

django/library/serializers.py
alee: sounds good - sorry for the scope creep on this PR :sweat_smile:

sgfost: No prob, can just tack on "contributors" in the title somewhere lol. Though I will say this serializer issue (or rather fixing it the right way) is a sort of a whole can of worms but one that should get dealt with so I don't mind adding it on the end here if you don't

alee: Maybe we should leave the serializer work off this PR and do it on a new branch then?

sgfost: Yeah that's best, I'm running into a roadblock now with some of the cleanup I was trying to do that's starting to cross over to comses/planning#136 territory

PR #662, opened 2023-08-02
feat: entity resolution / clustering for tags

hwelsters:
Attempts to close comses/planning#125

This branch was fixed since the previous one was a bit messed up. I fixed those mistakes and decided to use my own fork this time.

## Summary

Perform tag clustering and gazetteering with dedupe.

## Features
Tag Clustering takes in a list of tags, clusters the tags and simply picks the first tag in the cluster as the canonical tag.
Tag Gazetteer on the other hand allows the user to search for canonical tags that closely match the new tag from a canonical list and returns the canonical tags that most closely match the new tag. Furthermore, a confidence score for each suggestion is returned, determining how confident the model is that the new tag matches the canonical tag.

Tag Clustering is really only needed for creating the initial canonical list. Besides that, Tag Gazetteering will be used especially when adding new tags. If there is some initial canonical list, no matter how small, Tag Clustering will not be needed

### Curator commands
Created two new commands, one for clustering tags, the other for gazetteering / canonicalization.

To cluster:
```
./manage.py curator_cluster_tags
```
This command has the following optional arguments.
`--run` - Runs the clustering model, there are cases in which the curator might not want to run the model so this argument can be omitted,
`--label` - Lets the curator label the training data via the console.
`--save` - Actually save the results to the database. Without this argument, nothing will be saved.
`--threshold=[number from 0-1]` - This can be changed to alter the model threshold.

  
```
./manage.py curator_gazetteer_tags
```
This command has the following optional arguments.
`--run` - Runs the gazetteering model, there are cases in which the curator might not want to run the model so this argument can be omitted,
`--label` - Lets the curator label the training data via the console.
`--threshold=[number from 0-1]` - This can be changed to alter the model threshold.

## Tests
Wrote tests using Django tests


django/curator/tests/test_tag_deduplication.py
```diff
@@ -0,0 +1,117 @@
...
+            "Failed to split by ','",
+        )
+        self.assertEqual(list, type(preprocessed_Tags))
+
+
+class TestTagClustering(TestCase):
+    def setUp(self):
+        for i in range(300):
+            id = TestTagClustering._id_generator()
+            CanonicalTag(name=id).save()
```
alee, line 47: this would be more efficient if you create these objects directly and use [bulk_create](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#bulk-create) to persist them to the database

also, `id` is a [builtin Python function](https://docs.python.org/3/library/functions.html) and shadowing these (or any other standard library modules) is generally not recommended - that's why I changed the other one to `_id` but I missed this section of code

django/curator/tests/test_tag_deduplication.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def test_uncertain_pairs(self):
+        tag_clustering = TagClustering()
+        uncertain_pairs = tag_clustering.uncertain_pairs()
+        self.assertEqual(list, type(uncertain_pairs))
+        self.assertEqual(1, len(uncertain_pairs))
+        self.assertEqual(2, len(uncertain_pairs[0]))
+        self.assertEqual(int, type(uncertain_pairs[0][0]["id"]))
+        self.assertEqual(str, type(uncertain_pairs[0][0]["name"]))
+        self.assertEqual(str, type(uncertain_pairs[0][0]["slug"]))
```
alee, line 60: are there any assertions you can make about the contents of these data structures in addition to their types and shapes?

django/curator/tests/test_tag_deduplication.py
```diff
@@ -0,0 +1,117 @@
...
+            self.assertEqual(tuple, type(cluster[1]))
+
+    def _search(self):
+        tag_clustering = TagGazetteering()
+        for i in range(300):
+            uncertain_pair = tag_clustering.uncertain_pairs()
+            tag_clustering.mark_pairs(uncertain_pair, i % 3 != 0)
+        return tag_clustering.search({1: {"id": 1, "name": "abms"}})
+
+    def _id_generator(size=6, chars=string.ascii_uppercase + string.digits):
```
alee, line 116: code duplication with the other `_id_generator` in `TagPreProcessingTestCase`, maybe we should move it up as a top-level function declaration used by both.

django/curator/tag_deduplication.py
```diff
@@ -0,0 +1,162 @@
...
+        return os.path.exists(TagClustering.TRAINING_FILE)
+
+    def remove_training_file(self):
+        if os.path.isfile(TagClustering.TRAINING_FILE):
+            os.remove(TagClustering.TRAINING_FILE)
+
+
+class TagGazetteering(TagDeduplication):
+    TRAINING_FILE = "curator/gazetteering_training.json"
+    SEARCH_THRESHOLD = 0.5
```
alee, line 111: This should probably be parameterizable, with a django settings defined `DEFAULT_TAG_GAZETTEERING_THRESHOLD` that can be overridden when you create an instance, e.g., `TagGazetteering(search_threshold=0.8)`   

```
from django.conf import settings

...

DEFAULT_SEARCH_THRESHOLD = settings.TAG_GAZETTEERING_THRESHOLD
```

django/curator/tag_deduplication.py
```diff
@@ -0,0 +1,162 @@
...
+    def console_label(self):
+        dedupe.console_label(self.deduper)
+
+
+# This class is used to help to make the initial canonical list.
+# Besides that, there isn't much other use for this.
+# If the curator can directly provide a small canonical list, even a small one will do, this class will not be needed.
+class TagClustering(TagDeduplication):
+    TRAINING_FILE = "curator/clustering_training.json"
+    CLUSTERING_THRESHOLD = 0.1
```
alee, line 54: make this parameterizable with defaults defined in django settings (similar to SEARCH_THRESHOLD below)

django/curator/tag_deduplication.py
```diff
@@ -0,0 +1,162 @@
...
+from typing import List
+
+import dedupe
+from django.utils.text import slugify
+from taggit.models import Tag
+
+from curator.models import CanonicalTagMapping, CanonicalTag
+
+
+class TagPreprocessing:
```
alee, line 13: in general class names are nouns, not verbs. So this would be better named `TagPreprocessor`

Similar things apply to all the other class names

`AbstractTagDeduper` or `BaseTagDeduper`

`TagClusterer`

`TagGazetteer`

alee (changes requested):
Looks great, thanks for cleaning this up @hwelsters ! Some pedantic naming feedback included inline.

django/curator/tag_deduplication.py
sgfost: This seems unused as well, should it be?

django/curator/tag_deduplication.py
```diff
@@ -0,0 +1,161 @@
...
+        self.prepare_training()
+
+    # The training data is stored in a file.
+    # If it exists, load from the file
+    # Otherwise, start from scratch
+    def prepare_training(self):
+        data = self.prepare_training_data()
+        if os.path.exists(TagClusterer.TRAINING_FILE):
+            with open(TagClusterer.TRAINING_FILE, "r") as training_file:
+                self.deduper.prepare_training(data, training_file)
```
sgfost, line 65: Would it be better to store training data in the db instead of a file in the src tree? Looks like `Dedupe.prepare_training()` needs a file but this could be done by writing to an in-memory file object: https://docs.python.org/3/library/io.html#io.StringIO

django/curator/tag_deduplication.py
hwelsters: TagPreprocessor is supposed to be used as an upstream task. It is currently not used yet, I must have accidentally included it.

sgfost: Ah ok, if its meant to be used then it should be fine to include it

django/curator/tag_deduplication.py
hwelsters: I am not too sure what the benefits/costs are for storing the training data in the db vs file system but maybe I can make a function for it in case it is needed?

django/curator/management/commands/curator_cluster_tags.py
```diff
@@ -0,0 +1,59 @@
...
+        parser.add_argument(
+            "--threshold",
+            "-t",
+            help="Threshold to use for clustering tags",
+            action="store_true",
+            default=0.5,
+        )
+
+    def handle(self, *args, **options):
+        tag_gazetteering = TagClusterer(clustering_threshold=options["threshold"])
```
alee, line 45: add heredocs to this method describing how to use this management command to bootstrap tag clusters properly and then how to transition to `curator_gazetteer_tags`.

naming a TagClusterer instance `tag_gazetteering` seems copypasta wrong.

django/curator/management/commands/curator_cluster_tags.py
```diff
@@ -0,0 +1,59 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from curator.tag_deduplication import TagClusterer
+
+
+class Command(BaseCommand):
+    help = "Cluster Tags using dedupe"
```
alee, line 9: this help method should include the usage details you added to this PR, with clear and consistent language. How and when should this class be used, with example invocations and clear instruction on where things will be persisted, and how to parameterize / change that.

alee (changes requested):
Documentation needs to address how a specific tag in a cluster gets selected to be the `CanonicalTag` for that cluster.

django/curator/tag_deduplication.py
alee: I'm OK with it living on the filesystem. It's essentially private / transient data that is only used once to bootstrap the gazetteering if I understand the process correctly...

alee:
superseded by #671

PR #663, opened 2023-08-02
Refactor core/home app views and serializers

### LINKED ISSUE #490: noscript javascript detection isn't working
I'm not sure if the code here is supposed to be working, but it doesn't appear to be doing anything.

https://github.com/comses/comses.net/tree/master/frontend/public

We should be detecting when javascript is disabled and display a message on the page that some features of comses.net will be degraded without javascript

sgfost:
- merges related user/profile serializers to standardize on one minimal profile data type
- merges two vue components that both implemented a multiselect search over profiles
- also sneaks in a noscript tag in the base template for warning visitors with js disabled
- moves views/serializers for memberprofile, events, jobs, etc. so they are all in the same app as their model - comses/planning#136
--- 
- [x] resolves #490
- [x] resolves comses/planning#137


sgfost:
The intent here was to also do some re-organizing of django apps for comses/planning#136, though I'd like to be more confident in where things should end up before hacking away it it, so those changes are left out for now.

Current thought is to move everything `MemberProfile` to core and events/jobs/etc to home. The member profile serializer's dependency on library for related codebases was removed here already since that was unused and should probably be retrieved differently if it were to be needed.

alee:
is it easier to move events/jobs into home or the events/jobs templates into core?

sgfost:
My thought with moving events/jobs/etc. to home was to separate them from core functionality that can be used by all other apps so that its "beside" library. Maybe a new app that encapsulates ''non-library content" and use home only for pages and such?

alee:
That's a fair assessment, so home is wagtail related models + basic content related things and core is the shared common-to-all things models + views?

sgfost:
Yep, only issue would be that it adds a few more models/views to the already somewhat large home app but we can always think about a split later on

sgfost:
I'm actually going to walk back wanting to move events and jobs to home, creating and running safe migrations for moving models between apps is a bit of a nightmare and in this case would just net more confusion/potential for problems. There is at least one active PR to django for fixing this so I'd say wait until that makes it into a release to mess around with moving model tables

https://code.djangoproject.com/ticket/24686

Still leaves re-aligning views and serializers though, which is the majority of the refactor anyways

django/core/views.py
```diff
@@ -358,109 +265,310 @@ def discourse_sso(request):
...
+        else:
+            return self.queryset.with_user()
 
     def get_retrieve_context(self, instance):
-        context = {self.context_object_name: instance}
+        context = super().get_retrieve_context(instance)
+        accessing_user = self.request.user
+        # FIXME: ideally this functionality should be in the library app, though its not clear
+        # whether this can be done without a much more complicated workaround or a major
+        # re-organization of apps
```
sgfost, line 345: This is the only bit of weirdness left (importing from `library` in `core`). Couldn't come up with a clean solution for it but there must be one..

django/core/views.py
```diff
@@ -358,109 +265,310 @@ def discourse_sso(request):
...
+        logger.debug("Finding models for user %s", instance.user)
+        context["codebases"] = (
+            Codebase.objects.accessible(accessing_user)
+            .filter_by_contributor(instance.user)
+            .with_tags()
+            .with_featured_images()
```
sgfost, line 351: This is the only bit of weirdness left (importing from `library` in `core`). Couldn't come up with a clean solution for it but there must be one..

django/core/urls.py
```diff
@@ -42,23 +106,12 @@
...
-
-if not settings.DEPLOY_ENVIRONMENT.is_production:
-    import debug_toolbar
-
-    urlpatterns = [
-        path("argh/", handler500, name="error"),
-        path("make-error/", views.make_error),
-        path("__debug__/", include(debug_toolbar.urls)),
-    ] + urlpatterns
+urlpatterns += get_dev_urls()
```
alee, line 113: I'm pretty sure this will throw an error in prod e.g., `[1, 2, 3] + None`. Would need to fix `get_dev_urls` and `get_debug_urls` to `return []` in the else clause

It might be cleaner / easier to see what's happening by keeping the conditional guard where we mutate urlpatterns but don't feel too strongly about it either way

django/core/urls.py
```diff
@@ -25,12 +26,75 @@
 
 app_name = "core"
 
+
+def get_debug_urls():
+    if not settings.DEPLOY_ENVIRONMENT.is_production:
```
alee, line 31: see below but probably need to `return []` here

django/library/serializers.py
```diff
@@ -62,30 +65,30 @@ class Meta:
...
-            return (contributor.user if contributor else None, contributor)
+            user = contributor.user if contributor else None
         else:
-            # otherwise, match by the exact name given and blank email
             contrib_filter = {
                 k: validated_data.get(k, "")
                 for k in ["given_name", "family_name", "email"]
             }
-            return None, Contributor.objects.filter(**contrib_filter).first()
+            contributor = Contributor.objects.filter(**contrib_filter).first()
```
alee, line 89: does this clause also need to set `user` to None or is it OK for it to be the result of `validated_data.get("user") at the top of the method`?

alee (changes requested):
great refactor! Really like the winnowing down of duplicate types both in the frontend and backend serializers. Some minor comments in the notes though I think there might be some bugs in the conditional logic that aren't getting exposed yet.

If possible it would be great to write test(s) that do expose those pathways first to make the build fail and then the fix...

sgfost:
> If possible it would be great to write test(s) that do expose those pathways first to make the build fail and then the fix...

I'm assuming you're referring to the contributor serializer?

django/core/urls.py
sgfost: Thanks for catching that. Conditionals at the bottom is a bit more concise and maybe easier to follow, I'll go with that

django/library/serializers.py
sgfost: It may work, but you're right, much better to be explicit

django/core/views.py
alee: I think this about as best as we can make it for balancing simplicity + understandability

In the future if Django makes it easier to move models across apps it might be better for `core.MemberProfile` and its associated views / urls / templates to move to `home.MemberProfile` so that the `core` doesn't have any dependencies on other apps, while allowing other apps to have side linkages with each other as necessary

alee (approved):

PR #666, opened 2023-08-09
feat: Machine Leaning Spam Detection

BlllueSea:


django/curator/models.py
```diff
@@ -300,3 +309,45 @@ def migrate(self, new_names, old_name):
...
+            str(self.user_classifier_confidence),
+            str(self.labelled_by_curator),
+            str(self.last_updated),
+            str(self.is_training_data),
+        )
+
+
+# Create a new UserSpamStatus whenever a new MemberProfile is created
+@receiver(post_save, sender=MemberProfile)
+def sync_member_profile_spam(sender, instance, **kwargs):
```
sgfost, line 352: I believe we need to check the `created` argument https://docs.djangoproject.com/en/4.2/ref/signals/#post-save otherwise this will try and fail to create a second object when a memberprofile gets updated

django/curator/management/commands/curator_spam_detection.py
```diff
@@ -0,0 +1,135 @@
...
+
+class Command(BaseCommand):
+    help = "Perform spam detection"
+
+    def __init__(self):
+        self.detection = SpamDetection()
+        self.processor = self.detection.processor
+        self.user_meta_classifier = self.detection.user_meta_classifier
+        self.text_classifier = self.detection.text_classifier
+
```
sgfost, line 20: I think this may be better done with subcommands/subparsers (https://docs.python.org/3/library/argparse.html#sub-commands) since it seems the intent is to just run one action. Also may want to consider combining common workflows

django/curator/management/commands/curator_spam_detection.py
```diff
@@ -0,0 +1,135 @@
...
+        model_metrics = options["get_model_metrics"]
+        load = options["load_labels"]
+        train_user = options["train_user"]
+        predict_user = options["predict_user"]
+        p_train_user = options["p_train_user"]
+        train_text = options["train_text"]
+        predict_text = options["predict_text"]
+
+        load_directory = pathlib.Path(DATASET_FILE_PATH)
+
```
sgfost, line 117: could potentially shorten this to use dynamic method calls e.g. `getattr(self, f"handle_{action})()"`

alee:
apologies for the pivot, but after starting to go through the https://course.fast.ai I'm leaning towards using fast.ai for much of this (and pytorch as well) instead of tensorflow + scikit.

django/curator/management/commands/curator_spam_detection.py
BlllueSea: I'm still trying to figure out what to fix for this suggestion. I'll ask you again today at our meeting.

PR #667, opened 2023-08-18
Peer review: draft workflow, closing, invitation expirations

sgfost:
#### Major changes
- Force the peer review process to take place on unpublished releases by either creating a new copy draft release or converting an existing draft release into an `under_review` release
  - (resolves comses/planning#72)
- add `closed` to peer reviews so they can be marked as inactive by the submitter or editor
  - (resolves comses/planning#32)

#### other bugfixes/refactors:
- fix and refactor pr dashboard filtering
  - (resolves comses/planning#172)
- refresh expiration date when re-sending an invite
  - (resolves comses/planning#141)
- use the 'editing' editor when logging feedback actions instead of the editor stored on the invitation
  - (resolves comses/planning#176)


sgfost:
slightly underestimated the amt of consideration this needed. Tests are being worked on and will need to be somewhat thoroughly tested on staging as well

sgfost:
shouldn't provide an option for creating a new release when requesting from a draft -- only convert to 'under review'

sgfost:
this seems all be working correctly now, just need to decide on what to do with a release after peer review completes.

Current strategy is to automatically publish if the codebase has `live == True` (codebase has at least 1 other release that is published), otherwise set to `unpublished` which is neither a draft nor live and prevents editing files but remains private until published

alee:
I think the strategy of keep it unpublished if unpublished, publish if there's already a live one is a reasonable one. It might be safest though to make all steps explicit, so we keep it unpublished until they explicitly publish it (e.g., the "congratulations, peer review passed" email might have a direct link to publish the model)

sgfost:
Yeah good idea, that reminds me, I should add a link to edit the release in the email that gets sent to the submitter after changes are requested

frontend/src/components/releaseEditor/ReviewModal.vue
```diff
@@ -13,6 +13,7 @@
     >
       <template #body>
         <div class="alert alert-danger mb-4" role="alert">
+          <!-- FIXME: is there a way to deduplicate this + reminders.jinja -->
```
sgfost, line 16: easiest way would be mounting a vue component in place of the template. Feels silly doing so for static markup but may beat duplicating it

alee: I suspected that might be the only way forward since we regularly embed frontend things into the backend but not vice versa easily.. I think it's not a blocker for this PR moving forward, am OK with having two sets of guidance, I'll make an issue out of it maybe for a student dev to take on

django/library/models.py
```diff
@@ -1014,17 +1085,33 @@ class CodebaseRelease(index.Indexed, ClusterableModel):
     * git repository in /repository/<codebase_identifier>/
     """
 
+    class Status(models.TextChoices):
```
alee, line 1088: good work unifying liveness, draft, and peer review status here

frontend/src/stores/releaseEditor.ts
```diff
@@ -90,7 +91,7 @@ export const useReleaseEditorStore = defineStore("releaseEditor", () => {
     isInitialized.value = false;
     await fetchCodebaseRelease(identifier, versionNumber);
     await fetchMediaFiles();
-    if (!release.value.live) {
+    if (release.value.canEditOriginals) {
```
alee, line 94: :+1:

alee (approved):

PR #670, opened 2023-09-05
build(deps): upgrade to django 4.2 LTS

alee:
closes https://github.com/comses/planning/issues/151 
closes https://github.com/comses/planning/issues/154


django/core/models.py
```diff
@@ -556,6 +557,7 @@ def public(self):
 
 @add_to_comses_permission_whitelist
 class Event(index.Indexed, ClusterableModel):
+    EXPIRED_EVENT_DAYS_THRESHOLD = 2
```
sgfost, line 560: The constant/setting that does the same thing for jobs (`POST_DATE_DAYS_AGO_THRESHOLD`) is in `settings` (comes from `config.ini`), worth matching where these live? Probably naming too, something like `EXPIRED_JOB_DAYS_THRESHOLD`?

PR #671, opened 2023-09-11
feat: entity resolution

hwelsters:
Attempts to close comses/planning#125

## Summary

Perform tag clustering and gazetteering with dedupe.

## Features

Tag Clustering is needed for creating the initial canonical list. 

### Curator commands
Created four new commands, one for clustering tags, the other for gazetteering / canonicalization.
## 1 ```curator_cluster_tags``` 
This creates TagCluster objects. These can then be edited by going to curator_edit_clusters
- --label - Lets the curator label the training data via the console.
- --reset - The user is usually stopped from clustering if there is a clustering session that is not complete. This argument removes all clusters before clustering.
- --threshold=[number from 0-1] - This can be changed to alter the model threshold.
 
## 2 ```curator_edit_clusters``` 
This command lets the user edit clusters and then save the mappings to the database.
While modifying clusters, there are four options.
`(c)hange canonical tag name` - Lets you change the name of the canonical tag
`(a)dd tags` - Lets you add tags to the cluster
`(r)emove tags` - Lets you remove tags from the cluster
`(s)ave` - Saves the cluster to the database.
`(f)inish` - This does not save the cluster. It just means you are done with changing it and you're moving on. I decided not to autosave since there are certain cases where the user might want to just get rid of the cluster instead of saving it.
  
## 3 ```curator_map_tags``` 
This command attempts to map a tag to a canonical tag if it currently isn't already mapped.
- --label - Lets the curator label the training data via the console.
- --threshold=[number from 0-1] - This can be changed to alter the model threshold.
  
## 4 ```curator_modify_cannon``` 
This command is used if the user would like to modify the canonical list.
  
## Tests
Wrote tests using Django tests


sgfost:
Looks like migrations might be out of date, django generated an additional one when building

```
    operations = [
        migrations.RemoveField(
            model_name='canonicaltagmapping',
            name='is_canonical',
        ),
        migrations.AlterField(
            model_name='canonicaltagmapping',
            name='date_created',
            field=models.DateTimeField(auto_now=True),
        ),
        migrations.CreateModel(
            name='TagCluster',
            fields=[
                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                ('canonical_tag_name', models.TextField()),
                ('confidence_score', models.FloatField()),
                ('date_created', models.DateTimeField(auto_now_add=True, null=True)),
                ('tags', models.ManyToManyField(to='taggit.Tag')),
            ],
        ),
    ]
```

hwelsters:
Superseded by #673

PR #676, opened 2023-10-13
feat: add rich previews

CharlesSheelam:
Utilized the open graph protocol to add rich previews for Base page, codebases, events, and jobs.


django/core/jinja2/base.jinja
```diff
@@ -288,6 +288,21 @@
     <link rel="mask-icon" href="{{ static('favicon/safari-pinned-tab.svg') }}" color="#5bbad5">
     <meta name="msapplication-TileColor" content="#da532c">
     <meta name="theme-color" content="#ffffff">
+    {% block ogp_tags %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="https://www.comses.net/">
+    <meta property="og:type" content="website">
+    <meta property="og:title" content="CoMSES Net">
+    <meta property="og:description" content="CoMSES Net is an international open research community dedicated to fostering good practices for computational / agent based modeling.">
+    <meta property="og:image" content="https://complexity.asu.edu/sites/default/files/comses.jpg">
```
alee, line 297: these probably should refer to an internal static image.. I guess the current logo is generated by JS but it could live in core/static/images as well and be accessible via something like `{{ static("images/comses-logo.png") }}`. 

This can be done by copying `frontend/src/assets/images/comses-logo.png` to `django/core/static/images` and then replace all `og|twitter:image` references to the static invocation above

django/core/jinja2/base.jinja
```diff
@@ -288,6 +288,21 @@
     <link rel="mask-icon" href="{{ static('favicon/safari-pinned-tab.svg') }}" color="#5bbad5">
     <meta name="msapplication-TileColor" content="#da532c">
     <meta name="theme-color" content="#ffffff">
+    {% block ogp_tags %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="https://www.comses.net/">
+    <meta property="og:type" content="website">
+    <meta property="og:title" content="CoMSES Net">
```
alee, line 295: should we include the full expanded acronym here, Network for Computational Modeling in the Social and Ecological Sciences or will it be too long for the rich preview?

django/core/jinja2/core/jobs/list.jinja
```diff
@@ -1,6 +1,22 @@
 {% extends "search_layout.jinja" %}
 {% from "common.jinja" import list_page, paginator, breadcrumb, member_profile_href, search_tag_href, search_bar %}
 
+{% block ogp_tags %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="https://www.comses.net/jobs/">
+    <meta property="og:type" content="website">
+    <meta property="og:title" content="Community Jobs">
+    <meta property="og:description" content="View all jobs related to CoMSES">
+    <meta property="og:image" content="https://complexity.asu.edu/sites/default/files/comses.jpg">
```
alee, line 10: should we create custom images for comses jobs / events / codebases and add them to static? might be a nice UX touch

django/library/jinja2/library/codebases/releases/retrieve.jinja
```diff
@@ -56,6 +56,22 @@
     ]) }}
 {% endblock %}
 
+{% block ogp_tags %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="{{ request.build_absolute_uri(absolute_url) }}">
+    <meta property="og:type" content="article">
+    <meta property="og:title" content="{{ codebase.title }}">
+    <meta property="og:description" content="{{ codebase.description.raw|truncate(150) }}">
```
alee, line 64: we should probably use `codebase.summarized_description` (and modify its implementation to make it work better for this use case as needed)

django/core/jinja2/core/member_profiles/retrieve.jinja
```diff
@@ -14,6 +14,23 @@
     ]) }}
 {% endblock %}
 
+{% block ogp_tags %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="{{ request.build_absolute_uri(absolute_url) }}">
+    <meta property="og:type" content="article">
+    <meta property="og:title" content="{{ title }}">
+    <meta property="og:description" content="{{ description }}">
```
alee, line 22: I think this is a copypasta issue, for MemberProfile this should probably be `name` and `bio` ?

django/core/jinja2/core/jobs/retrieve.jinja
```diff
@@ -16,6 +16,22 @@
     ]) }}
 {% endblock %}
 
+{% block ogp_tags %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="{{ request.build_absolute_uri(absolute_url) }}">
+    <meta property="og:type" content="article">
+    <meta property="og:title" content="{{ title }}">
+    <meta property="og:description" content="{{ description }}">
```
alee, line 24: probably best to use `summary` instead, as description can be quite long

same thing with Events

alee (changes requested):
Good work @CharlesSheelam ! This sets up the groundwork nicely. A few minor nitpicks and suggested edits included inline.

We should also identify other "important" pages to put these OGP tags in as well...

- education
- resources/modeling-frameworks (https://www.comses.net/resources/modeling-frameworks/)
- documentation
- standards
- ?

django/core/jinja2/base.jinja
CharlesSheelam: We could test it the full expanded acronym, I have changed it to the full thing for now. If it seems to long later I can change it back.

django/core/jinja2/base.jinja
```diff
@@ -288,6 +288,21 @@
     <link rel="mask-icon" href="{{ static('favicon/safari-pinned-tab.svg') }}" color="#5bbad5">
     <meta name="msapplication-TileColor" content="#da532c">
     <meta name="theme-color" content="#ffffff">
+    {% block ogp_tags %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="https://www.comses.net/">
+    <meta property="og:type" content="website">
+    <meta property="og:title" content="The Network for Computational Modeling in the Social and Ecological Sciences (CoMSES Net)">
+    <meta property="og:description" content="CoMSES Net is an international open research community dedicated to fostering good practices for computational / agent based modeling.">
+    <meta property="og:image" content="/static/images/logo.png">
```
alee, line 297: refer to static assets in a jinja template using the incantation I shared earlier,

`{{ static("images/logo-comses.png") }}`

so the full line would be something like

```
<meta property="og:image" content="{{ static('images/logo-comses.png') }}">
```

For more info on why we dynamically generate these static links, 

https://docs.djangoproject.com/en/4.2/howto/static-files/#configuring-static-files

(though the syntax looks a little different since we are using Jinja templates instead of Django templates)

alee (changes requested):
LGTM! Just a minor note to replace the og:image /static/image URL with the jinja `static()` function instead

django/core/jinja2/base.jinja
CharlesSheelam: Apologies, forgot to fix these. I just updated them and the pull request is updated.

PR #681, opened 2023-11-06
fix: display correct image for codebases in ogp tags

CharlesSheelam:
- Fixed the image tags in `django/library/jinja2/library/codebases/releases/retrieve.jinja` to display the correct image.

- Fixed image tag in `django/core/jinja2/core/jobs/retrieve.jinja` to the correct syntax.

### DIFF
--- a/django/core/jinja2/base.jinja
+++ b/django/core/jinja2/base.jinja
@@ -1,4 +1,4 @@
-{% from "common.jinja" import ogp_tags_base%}
+{% from "common.jinja" import render_ogp_tags%}
 
 {% if not user %}
     {% set user = request.user %}
@@ -291,7 +291,7 @@
     <meta name="msapplication-TileColor" content="#da532c">
     <meta name="theme-color" content="#ffffff">
     {% block ogp_tags %}
-    {{ ogp_tags_base() }}
+    {{ render_ogp_tags() }}
     {% endblock ogp_tags %}
     <title>
         {% block title %}CoMSES Net{% endblock %}
--- a/django/core/jinja2/base.jinja
+++ b/django/core/jinja2/base.jinja
@@ -272,6 +272,23 @@
     {% endwith %}
 {% endmacro %}
 
+{% macro ogp_tags_base(url=None, title=None, description=None, image=None) %}
+    <!-- Facebook Meta Tags -->
+    <meta property="og:url" content="{{ url or 'https://www.comses.net/'}}">
+    <meta property="og:type" content="website">
+    <meta property="og:title" content="{{ title or 'The Network for Computational Modeling in the Social and Ecological Sciences (CoMSES Net)'}}">
+    <meta property="og:description" content="{{ description or 'CoMSES Net is an international open research community dedicated to fostering good practices for computational / agent based modeling.'}}">
+    <meta property="og:image" content="{{ image or static('images/logo-comses.png') }}">
+
+    <!-- Twitter Meta Tags -->
+    <meta name="twitter:card" content="summary_large_image">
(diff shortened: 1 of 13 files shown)


django/library/jinja2/library/codebases/releases/retrieve.jinja
```diff
@@ -62,16 +62,16 @@
     <meta property="og:type" content="article">
     <meta property="og:title" content="{{ codebase.title }}">
     <meta property="og:description" content="{{ codebase.summarized_description }}">
-    <meta property="og:image" content="{{ codebase.get_featured_image() }}">
+    <meta property="og:image" content="{{ codebase.get_image_urls()[0] }}">
```
alee, line 65: do we need to guard against codebases with no images / empty `get_image_urls()` ?

it might be useful to use wagtail's image processing to generate a specifically sized rendition and a URL to that rendition:

https://stackoverflow.com/questions/71509212/getting-url-of-image-on-django-wagtail-site

https://docs.wagtail.org/en/stable/topics/images.html#generating-the-image-as-foo-to-access-individual-properties

https://docs.wagtail.org/en/stable/topics/images.html#full_url

sgfost:
also just realized the tags are duplicated across each page, using a macro https://jinja.palletsprojects.com/en/3.1.x/templates/#macros with the relevant params (title, description, optional image url) would cut down on quite a bit of markup

django/core/jinja2/base.jinja
```diff
@@ -272,6 +272,23 @@
...
+    <meta property="og:image" content="{{ image or static('images/logo-comses.png') }}">
+
+    <!-- Twitter Meta Tags -->
+    <meta name="twitter:card" content="summary_large_image">
+    <meta property="twitter:domain" content="comses.net">
+    <meta property="twitter:url" content="{{ url or 'https://www.comses.net/'}}">
+    <meta name="twitter:title" content="{{ title or 'The Network for Computational Modeling in the Social and Ecological Sciences (CoMSES Net)'}}">
+    <meta name="twitter:description" content="{{ description or 'CoMSES Net is an international open research community dedicated to fostering good practices for computational / agent based modeling.'}}">
+    <meta property="og:image" content="{{ image or static('images/logo-comses.png') }}">
+{% endmacro %}
```
sgfost, line 290: I think this should go in core/common.jinja with most of the other shared macros and then be explicitly imported where we use it

django/library/jinja2/library/codebases/releases/retrieve.jinja
```diff
@@ -62,14 +62,24 @@
...
+        <meta property="og:image" content="{{ static('images/logo-comses.png') }}">
+    {% endif %}
     <!-- Twitter Meta Tags -->
     <meta name="twitter:card" content="summary_large_image">
     <meta property="twitter:domain" content="comses.net">
     <meta property="twitter:url" content="{{ request.build_absolute_uri(absolute_url) }}">
     <meta name="twitter:title" content="{{ codebase.title }}">
     <meta name="twitter:description" content="{{ codebase.summarized_description }}">
-    <meta name="twitter:image" content="{{ codebase.get_featured_image() }}">
+    {% if codebase.featured_images.exists() %}
```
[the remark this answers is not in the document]
sgfost: Still the case, especially with a macro. Probably best to just modify get_featured_rendition_url() to return None if the featured image doesn't exist and then we'd just need one line for the template macro:

```
{{ ogp_tags(image=codebase.get_featured_rendition_url(), ...) }}
```

sgfost (changes requested):
Looks good, thanks Charles. 2 nitpicks and then this is good to go

PR #684, opened 2023-11-16
build(deps): upgrade to debian bookworm node-lts

alee:
some package in package.json requires a binary version of git for some reason and was causing `yarn upgrade` to fail. should investigate further...

```
root@24de4b0e2c6e:/code# yarn upgrade
yarn upgrade v1.22.19
[1/4] Resolving packages...
error Couldn't find the binary git
```

also attempting an upgrade to yarn modern: https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern


sgfost:
that seems like yarn itself can't find git, somehow..

alee:
right, maybe there's a git+ssh dependency somewhere in there?

I'm not sure why `yarn lint` is failing here, seems fine on my local system..

sgfost:
I just had that same problem upgrading stuff on pom, resolved itself by hard refreshing (deleting/reinstalling) yarn.lock

sgfost:
That seems like it did the trick (it just bumped eslint). `yarn upgrade` also works for me locally

alee:
> That seems like it did the trick (it just bumped eslint). `yarn upgrade` also works for me locally
> 
> I can push the new lockfile if it won't mess with your branch

Hmm, I rebuilt the lockfile and the container without git and it yarn still complains about git. Or are you talking about just fixing the `yarn lint` issue that's failing this build?

sgfost:
I meant the git binary issue didn't come up for me, though I'm realizing now I didn't actually make sure the container rebuilt fully

alee:
hmm, modern yarn seems to run the github action out of memory, need to investigate further when time permits

sgfost:
presumably yarn PnP uses way more memory than the runner has. I also can't manage to get the default PnP install strategy working at all with docker locally. Was that the intent with upgrading to yarn modern?

sgfost:
yarn seems to revert itself to v1 if we don't specify _and commit_ the binary

looks like corepack https://yarnpkg.com/corepack may be the way to lock the yarn vers. without needing to commit the whole thing

alee:
@sgfost any ideas why these tests are failing now? I didn't see anything in the wagtail 5.2.2 release notes, perhaps a different dependency upgrade? https://github.com/comses/comses.net/pull/684/commits/d960a22dc66a49a8c8cd1c3834a1a9fb325e5420

sgfost:
Seems to be failing regardless of any of those upgrades (including pinning django-modelcluster to 6.1). I'm going to try debugging the test but I have no idea how the issue could have suddenly arisen

sgfost:
Oh whoops.. y2k bug in those tests. Satisfying yet very silly conclusion

alee:
> Oh whoops.. y2k bug in those tests. Satisfying yet very silly conclusion

leap year bugs are the best

PR #696, opened 2024-02-15
build refactor: replace config.ini with environment variables (via single .env file) and fix templating issues

sgfost:
attempts to consolidate all app configuration into environment variables

.env is created from a template and generated secrets are placed in, all variables defined in .env are then fed into the server container. The django settings module then reads from os.getenv() rather than configparser

---

**list of changes**:
- removed `config.ini`/`config.ini.template`
- removed `docker.env`
- removed targets for any non-generated secrets, should be managed directly in the .env file
- removed pgpass.template in favor of just echoing the correct format
- set environment variables in server container from .env via `env_file` attribute
- add convenience script `envreplace` for setting values in a file with GNU or BSD sed

---

resolves comses/planning#145


alee:
I'd like to move towards something like https://docs.docker.com/compose/use-secrets/ - sorry for not including it in earlier issues / docs. I think we can pivot this PR into that work though

sgfost:
That should be straightforward to add. I hadn't already done so mostly because I wasn't entirely sure what the benefit of using secrets sans swarm-mode (files in a tmpfs mount, I think?) over environment variables was

alee:
I'm mostly going on their docs here:

> Docker Compose provides a way for you to use secrets without having to use environment variables to store information. If you’re injecting passwords and API keys as environment variables, you risk unintentional information exposure. Environment variables are often available to all processes, and it can be difficult to track access. They can also be printed in logs when debugging errors without your knowledge. Using secrets mitigates these risks.

The other thing though is to support granular secrets for each thing that needs the secret so that updating or adding new environment variables or secrets doesn't blow away the previous set

django/core/settings/defaults.py
```diff
@@ -231,50 +230,34 @@ def is_test(self):
 # from now to 90 days ago
 ADMIN_DASHBOARD_DAYS = 90
 
-config = configparser.ConfigParser()
-# FIXME: switch to docker secrets instead
-config.read("/run/secrets/config.ini")
-
-RELEASE_VERSION = config.get("default", "BUILD_ID", fallback="v2023.01")
+RELEASE_VERSION = os.getenv("RELEASE_VERSION", "v2024.01")
```
alee, line 233: need to make BUILD_ID match up with RELEASE_VERSION in .env.template

deploy/conf/.env.template
```diff
@@ -0,0 +1,55 @@
+# app
+BUILD_ID=
```
alee, line 2: rename to RELEASE_VERSION here and in the Makefile

PR #699, opened 2024-03-14
provide automated DOI registration

alee:
add support for automated metadata translation from our object model into the DataCite schema with eventual DOI publication via the datacite Python library

refs https://github.com/comses/planning/issues/146


django/library/models.py
```diff
@@ -2464,6 +2481,203 @@ def to_dict(self):
...
+        Important: this should be call AFTER the release has been published as DataCite
+        requires the copyrightYear information
+
+        See https://codemeta.github.io/crosswalk/datacite/ for documentation on CodeMeta to DataCite crosswalk.
+
+        We don't want to make a copy since some field names are different and also
+        some fields DataCite do not want or have.
+        """
+        metadata = {}
+        codemeta = release.codemeta.metadata
```
sgfost, line 2509: Do we want to go through codemeta at all for any of this? Seems like it may be easier to follow if we just transform straight from the release, and use shared methods for any instances where fields may have the same transformation as codemeta

monaw: originally we thought that all the info from codemeta will go straight into the datacite doi but found out that's not the case.  i can refactor to use the release info as much as possible instead the codemeta.  @alee , ok?

monaw: @sgfost - can you elaborate on the "shared methods" and how this is done in python syntax?  for example CodeMeta convert_authors method, how to share that with the DataCiteMetadata?

alee: @asuworks and I had a conversation about this yesterday - there's two ways we could think about this, one is that codemeta represents our "Intermediate Representation (IR)" metadata that we use to translate to all other metadata. That would be fine if there's a non-lossy transformation between our object model metadata -> codemeta -> datacite

However the issue is if there's additional / richer metadata fields in datacite than codemeta provides that we could use but aren't available so we need to directly use our object model metadata to transform into datacite, in that case it might be better to have a direct transformation from our object model metadata -> datacite

I think we did some analysis on this in the requirements spec in the metadata crosswalk table but don't have it clear in mind at the moment, we should look at that and respond definitively here if this is the case

alee: to answer your question though we could think of "shared methods" as functions that take our object model and convert it into a fully materialized in-memory intermediate representation (Python dictionary based or otherwise) that can be:

1. used to more easily transform into codemeta or datacite
2. easily tested without having to deal with making additional database queries, etc
3. maybe other benefits

asuworks: It would be nice to be able to do something like this: 

`
codebase_release = CodebaseReleaseFactory.create()

codemeta_transformer = CodeMetaTransformer(codebase_release)
codemeta_dict = codemeta_transformer.transform()

datacite_transformer = DataCiteTransformer(codebase_release)
datacite_dict = datacite_transformer.transform()
`

asuworks: Here is a code organization suggestion. @alee, @sgfost what do you think?
```python
from abc import ABC, abstractmethod

class SharedMetadataTransformer(ABC):
    SHARED_MEATADATA_SCHEMA = """schema definition..."""
    
    def __init__(self, codebase_release):
        self.codebase_release = codebase_release
    
    def transform(self):
        shared_metadata = {}
        shared_metadata.update(
            title = self.codebase_release["title"]
            authors = self.transform_authors()
            programming_languages = self.transform_programming_languages()
            # ... other attributes
            )
            
        self.validate(shared_metadata, self.SHARED_MEATADATA_SCHEMA)
        return shared_metadata
        
    def validate(self, metadata_dict, metadata_schema):
        jsonschema.validate(json.loads(metadata_dict), metadata_schema)
    
    # "shared methods"
    def transform_authors(self):
        print("Not implemented!")
        return self.codebase_release.contributors.all()
    
    def transform_programming_languages(self):
        return self.codebase_release.programming_languages.get()

class CodeMetaTransformer(SharedMetadataTransformer):
    CODEMETA_SCHEMA = """schema definition...."""
    
    def __init__(self, codebase_release):
        super().__init__(codebase_release)
    
    def transform(self):
        shared_metadata = super.transform()
        
        codemeta_data = {
            "@context": "http://schema.org",
            "@type": "SoftwareSourceCode",
            "shared_metadata_attribute": shared_metadata["shared_metadata_attribute"]
            "codemeta_specific_attribute": self.codebase_release.codemeta_specific_attribute,
            
            # Transform other attributes as needed...
        }
        
        super.validate(codemeta_data, self.CODEMETA_SCHEMA)
        return codemeta_data

class DataCiteTransformer(SharedMetadataTransformer):
    DATACITE_SCHEMA = """schema definition...."""
    
    def __init__(self, codebase_release):
        super().__init__(codebase_release)
    
    def transform(self):
        shared_metadata = super.transform()
        
        datacite_data = {
            "@context": "http://schema.org",
            "@type": "SoftwareSourceCode",
            "shared_metadata_attribute": shared_metadata["shared_metadata_attribute"]
            "datacite_specific_attribute": self.codebase_release.datacite_specific_attribute,
            
            # Transform other attributes as needed...
        }
        
        super.validate(datacite_data, self.DATACITE_SCHEMA)
        return datacite_data
```

sgfost: I like that, I think the `Transformer` naming makes sense. Any reason to not keep the `build()`/`transform()` as a `@classmethod` for the child classes? It makes the interface slightly simpler, not sure if there is any downside

Also, it may make more sense to pick and choose methods that the base transformer class provides, rather than using its `transform()` which is sorta just an arbitrary collection of fields that wouldn't mean much if we, say, needed a 3rd metadata crosswalk for something

asuworks: The reason to use a base class transform() is based on the assumption that it would generate an object which would cover (mostly) all the required fields in an acceptable format. 
In a child class we would just need to "fix" the wrong or missing attributes.
@sgfost can you provide a quick code sample of your idea?

monaw: since i'm new to our code and python, i'll defer the decision to Allen and Scott.  should i hold off on my refactor and work on the fabrica api until a decision is made or should i refactor using the CodebaseRelease info?

sgfost: something to this effect @asuworks, similar strategy just calling the methods from the base directly rather than building everything and picking out what we want from a dict
```python
class SharedMetadataTransformer:
    def __init__(self, release):
        self.release = release

    def convert_contrib_to_authors(self):
        # something
        pass

class CodemetaTransformer(SharedMetadataTransformer):
    INITIAL_DATA = {
            "@context": "http://schema.org",
            # ...
    }

    def transform(self):
        self.metadata = self.INITIAL_DATA.copy()
        self.metadata.update(
                # ...
                codemeta_specific="something"
                authors=self.convert_contrib_to_authors()
                # ...
        )
        return self.metadata
```



@monaw We'll still need to identify any shared transforms and remove the datacite class's dependency on codemeta regardless of how exactly it ends up looking. Focusing on the api client in the meantime isn't a bad idea though

alee: Nice ideas! I could see the use of having transform() generate a Python dictionary and return it every time, but we could also still do something like `DataCiteMetadata.build()` or `CodeMeta.build()` and use the transformer within it. One thing that might be good to do is to cache the Python dictionary that gets built up by the transformer, to have a more stateful object (though I certainly appreciate the functional programmingness of this design).

I tend to prefer composition over inheritance so might suggest that DataCiteMetadata and CodeMeta _have_ a MetadataTransformer instance that can convert a CodebaseRelease into a pure in-memory data structure that can then be used as faithful source material for further CodeMeta or DataCite transformations. The MetadataTransformer can then be mocked and tested without db dependencies etc.

Other thoughts?

django/library/doi.py
```diff
@@ -0,0 +1,953 @@
...
+        if invalid_releases:
+            logger.error(f"Failure. Metadata not in sync with DataCite for {len(invalid_releases)} releases: {invalid_releases}")
+        else:
+            logger.info(
+                f"Success. Metadata in sync with DataCite for all releases with DOI.")
+
+
+"""
+RECURRENT TASKS
+"""
```
sgfost, line 749: all of these should be available to be invoked from the management command. I also think the input() waits should be made optional so that we can schedule them as cron jobs for example

sgfost (commented):
Looks good, very thorough. Some thoughts from a first pass over everything.

also, I haven't fully combed through the code in all of the tasks in doi.py but I wonder if there any duplication between them that can be cleaned up?

monaw:
documentation @ https://docs.google.com/document/d/19-FdJcLdNYpMW4rgc85EzBq1xuI3HqZ3PVagJ-1nXlM/edit#heading=h.qu5a82ym9f0r

django/library/models.py
```diff
@@ -2438,14 +2510,90 @@ def __str__(self):
         return f"[peer review] {invitation.candidate_reviewer} submitted? {self.reviewer_submitted}, recommendation: {self.get_recommendation_display()}"
 
 
-class CodeMeta:
+class CommonMetadata:
```
sgfost, line 2513: it seems like this should be a collection of helper functions/static methods instead of a built dictionary, is this a worthwhile change?

django/library/models.py
```diff
@@ -2468,105 +2619,435 @@ def __init__(self, metadata: dict):
...
+        if not metadata:
+            raise ValueError(
+                "Initialize with a base dictionary with DataCite terms mapped to JSON-serializable values"
+            )
+        self.metadata = metadata
 
     @classmethod
-    def convert_platforms(cls, codebase_release: CodebaseRelease):
-        return [tag.name for tag in codebase_release.platform_tags.all()]
+    def from_codebase(cls, codebase: Codebase):
```
sgfost, line 2737: would it be better if `ReleaseDataciteMetadata` and `CodebaseDataciteMetadata` were 2 separate sibling classes that inherit from the same schema or at least the base fields that are currently duplicated?

sgfost (commented):
The core doi creation logic seems alright to me, the only real improvement I can think of other than minor cleanup is doing metadata updates on post_save signals rather than the complicated-seeming comparison. This should probably be done with celery though..

Otherwise, the only thing I'm confused by is the `remove_dois_from_not_peer_reviewed_releases` command. If the DOI exists and points to the release, is there harm in keeping it stored, even though its sorta 'legacy' data?

alee:
> The core doi creation logic seems alright to me, the only real improvement I can think of other than minor cleanup is doing metadata updates on post_save signals rather than the complicated-seeming comparison. This should probably be done with celery though..

Good call, though I think celery is probably overkill for this, a cron job should be fine since these aren't long-running or frequent processes. I don't think I would do it on `post_save`, and instead batch it nightly.

> Otherwise, the only thing I'm confused by is the `remove_dois_from_not_peer_reviewed_releases` command. If the DOI exists and points to the release, is there harm in keeping it stored, even though its sorta 'legacy' data?

I think the goal is to re-mint old handle.net and legacy PIDs to be uniform though @asuworks would probably be better to clarify here...

django/library/models.py
alee: One of the initial goals was to also provide an easier way to mock out the metadata tests and improve testing speed by not needing a DB but the design didn't end up that way. That's part of the reason that the CodeMeta and DataCite metadata classes were initialized with a bare dictionary but without validation it's kind of a mess. Let's discuss at or after tomorrow's dev meeting?

django/library/models.py
alee: That sounds like a good way to separate the logic, I'll work on that while continuing to clean up the way their dictionaries are built up

asuworks:
We wanted to get rid of legacy DOIs for consistency. There were 6 legacy DOIs in prod, as far as I remember... 
We would need to handle them separately. Do we want this?

sgfost:
Does this need any more work that isn't related to the metadata conversion code? Once this is merged I'll need to reconcile #718 so I think it makes sense to clean up the metadata system then

PR #700, opened 2024-03-15
File safety and digest page/ogp fixes

sgfost:
Putting a few quick things in here to get looked over @alee

a9e7b87 is mostly straightforward, need to check on staging whether it actually fixes the issue with images not working. Otherwise, its a matter of whether removing all of the storage paths from `.env` is the right move

### DIFF
--- a/django/core/jinja2/common.jinja
+++ b/django/core/jinja2/common.jinja
@@ -301,16 +301,26 @@ Currently in <em><mark>{{ constants.DEPLOY_ENVIRONMENT }}</mark></em> mode.
 {{ date.strftime('%B %d, %Y %I:%M %p') }}
 {% endmacro %}
 
-{% macro render_ogp_tags(url=None, title=None, description=None, image=None) %}
-    <!-- Open Graph Protocol Tags -->
-
+{%
+    macro render_ogp_tags(
+        request,
+        title="The Network for Computational Modeling in the Social and Ecological Sciences (CoMSES Net)",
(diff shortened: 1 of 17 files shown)


django/core/jinja2/common.jinja
```diff
@@ -301,16 +301,26 @@ Currently in <em><mark>{{ constants.DEPLOY_ENVIRONMENT }}</mark></em> mode.
...
+{%
+    macro render_ogp_tags(
+        request,
+        title="The Network for Computational Modeling in the Social and Ecological Sciences (CoMSES Net)",
+        description="CoMSES Net is an international open research community dedicated to fostering good practices for computational / agent based modeling.",
+        image=None
+    )
+%}
+    {% set url = request.build_absolute_uri() %}
+    {% set absolute_image_url = request.build_absolute_uri(image or static("images/logo-comses.png")) %}
```
alee, line 313: should come up with a better default image but that can be done later as well: 

https://mastodon.social/@comses

django/library/models.py
```diff
@@ -945,7 +945,7 @@ def create_release(
             release = self.create_release_from_source(source_release, release_metadata)
 
         if initialize:
-            release.get_fs_api().validate_bagit()
+            release.get_fs_api()  # implicitly initializes the release filesystem
```
alee, line 948: might create an `init_fs_api()` alias with `get_fs_api()`

alee (approved):
LGTM, thanks Scott!

PR #714, opened 2024-04-29
update digest archives

sgfost:
We still need a link to the digest page, I would say just stick it right here for now:
Screenshot from 2024-04-29 14-59-39
does that seem alright, @alee?


alee:
sounds good! would be nice to revamp how those secondary tab menus are set up; it's kind of a pain at the moment and involves a fair bit of duplication. I think how we were setting these up was as a management command since each individual page in the tab needs to have that secondary tab nav set on it

sgfost:
Yeah, I'd like to think through a bit of a navigation cleanup somewhat soon including comses/planning#114. I don't think the digest page needs to live in /about/ long-term so just a direct link in those tabs should suffice? I can put a script in here for doing so to version it

alee (approved):
LGTM, only residual thing that's a little weird is that when you click on the digest page itself you lose the nav that got you there but I don't think there's an easy way to set that up since it's a DigestView, not a legit wagtail Page

PR #693, opened 2024-01-12
Feat: Spam Detection Feature

aimura09:
Attempts to close https://github.com/comses/planning/issues/113

Squashed commits and solved merge conflicts.

## Summary
Management commands for Machine Learning spam detection.

## Features
Before running the commands, make sure spam_dataset.csv is located in the curator folder.
spam_dataset.csv consist of user_id and is_spam columns. is_spam column contains 1(Spam) or 0(Ham).

- XGBoostClassifier() ... Uses XGboost as a classifier.  Takes a data frame that has columns "user_id" and "input_data." The "input_data" column is a numerical vector where the selected fields are encoded by an encoder.

- CountVectEncoder() ... Uses CountVectorizer as an encoder. Takes selected fields from "user_id," "labelled_by_curator," "first_name," "last_name," "is_active," "email," "affiliations," "bio," "research_interests" of the MemberProfiles as input.

1. Run the following command to get a list of spam users.
    `./manage.py curator_spam_detection --predict`

2. options
    `./manage.py curator_spam_detection --fit`
    `./manage.py curator_spam_detection --get_model_metrics`
    `./manage.py curator_spam_detection --load_labels`

## Tests  
Wrote 16 unit tests using Django tests

PR #719, opened 2024-05-11
basic spam detection

sgfost:
uses two methods:
- timing form load->submission time
- using a honeypot field hidden from real users with css

logs actions to a SpamModeration model with a generic fk to content models. actions such as deactivating the user or allowing the post can then be taken via a WIP admin view

#### TODO:
- [x] test coverage
- [x] usable admin view with
  - confirming/rejecting spam assessment
  - filtering
  - option to deactivate user on confirm (or default?)
  - email reminder when X number of spam objects pile up, or after Y amt of time (?)
- [x] add more context to `SpamContent` detailing how/why we think its spam, flexible enough that it can also be used for detecting with NLP (think ahead to reconciling #693)

resolves comses/planning#235


sgfost:
spam content, and deleted events/jobs for that matter, still show up in the site-wide search. Need to find a good way to filter those from the general search but struggling to figure out how it currently does this for non-live codebases..

alee:
> spam content, and deleted events/jobs for that matter, still show up in the site-wide search. Need to find a good way to filter those from the general search but struggling to figure out how it currently does this for non-live codebases..

good catch, we should have a 
```
@classmethod
get_indexed_objects(cls)
```
defined for Events, MemberProfiles, Jobs

sgfost:
ah that's it, thanks

The bulk of this is done, just need to put up a few tests. I think it ought to work as a more flexible base for using the work in #693 as another detection method, but lmk if you have any thoughts or concerns when you get around to it

django/core/views.py
```diff
@@ -447,6 +450,12 @@ class EventViewSet(CommonViewSetMixin, OnlyObjectPermissionModelViewSet):
         "start_date",
     )
 
+    def get_queryset(self):
+        # exclude spam from list view
+        if self.action == "list":
+            return self.queryset.public()
```
alee, line 456: am I understanding this correctly to allow possible spam events to be accessible in the detail view? (same for Jobs?)

django/curator/views.py
```diff
@@ -1,14 +1,17 @@
 from django.contrib.auth.decorators import permission_required
+from django.contrib import messages
+from django.shortcuts import get_object_or_404, redirect
 from django.http import HttpResponseRedirect, HttpResponseBadRequest
+from django.urls import reverse
 from django.views.decorators.http import require_POST
-from wagtail.contrib.modeladmin.helpers import AdminURLHelper
+from wagtail_modeladmin.helpers import AdminURLHelper
```
alee, line 7: should eventually migrate to Snippets or alternative https://github.com/comses/planning/issues/248

django/library/jinja2/library/codebases/releases/edit.jinja
```diff
@@ -1,9 +1,13 @@
 {% extends "base.jinja" %}
+{% from "common.jinja" import spam_detected_alert %}
 
 {% block introduction %}
 <h1><i class='fas fa-cloud-upload-alt'></i> Publish your computational model</h1>
 {% endblock %}
 {% block content %}
+{% if release.codebase.is_marked_spam %}
+    {{ spam_detected_alert() }}
```
alee: Assume this should be `alert_if_spam(true)` ? and prevent release editing if marked as spam?

alee (approved):
LGTM thanks @sgfost ! I think I fixed a few minor issues with the codebase release edit form still including an outdated reference to the spam alert macro, will wait and see if things still look OK on staging before merging

django/core/jinja2/core/events/retrieve.jinja
```diff
@@ -21,6 +21,8 @@
 {% endblock ogp_tags %}
 
 {% block content %}
+    {{ alert_if_spam(is_marked_spam) }}
+    {{ alert_if_deleted(is_deleted, "event") }}
```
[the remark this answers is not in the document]
sgfost: Yeah, its now treated exactly the same as things marked spam, hidden from list view but accessible if you have a direct link to it. There wasn't much reasoning behind it besides simplifying queries

django/core/views.py
sgfost: Yes, mostly to give an explanation of what happened should a real post get mistakenly flagged. Though it should probably only be accessible by admins and the submitter

django/core/mixins.py
```diff
@@ -180,3 +184,112 @@ def list(self, request, *args, **kwargs):
 
         serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data)
+
+
+class SpamCatcherSerializerMixin(serializers.Serializer):
```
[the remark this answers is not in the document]
sgfost: Probably, yeah. the mixin shouldn't be a serializer. I'll check this out

sgfost:
Thx, I'll make a new issue with these needed changes as to not block merging if its in a 'good enough' state

PR #720, opened 2024-05-17
230 cml refactor contributor edit metadata form affiliations handling

asuworks:
## e2e (add some cypress tests later maybe?)

1. ADD CONTRIBUTOR 
   1. [X] From **Scratch (New Contributor Form)**
      1. [X] with a **unique email** (non existent in DB) 
         1. **assert frontend:**
            1. [X] all fields are correctly displayed in UI
            2. [X] all fields are enabled (editable)
         2. **assert backend:**
            1. [X] all fields are correctly stored in DB
            2. [X] update/edit is allowed
      2. with an **existing Contributor.email** (**Contributor is used in a release**) 
         1. **assert frontend:**
            1. [X] error message displayed: Contributor with this email already exsists!
         2. **assert backend:**
            1. [X] update/edit is not allowed: backend should not allow PUT request to `/contributors/` if a Contributor with the same email exists and is used by another release
      3. with an **existing Contributor.email** (**Contributor is NOT used in a release**) 
         1. **assert frontend:**
            1. [X] all fields are correctly displayed in UI
            2. [X] all fields are enabled (editable)
         2. **assert backend:**
            1. [X] existing Contributor fields will overwritten, since it is not used in any release 
            2. [X] all fields are correctly stored in DB
            3. [X] update/edit is allowed
      4. with an **existing AuthUser.email** \[NOT IMPLEMENTED\] [Current bahavior: a new Contributor is created and not linked to existing user]
         1. **assert frontend:**
            1. error message displayed: a User with this email exists, look it up!
         2. **assert backend**
            1. update/edit is not allowed
   2. From **existing Contributor**
      1. **Contributor is USED in at least in ONE other release**
         1. **assert frontend:**
            1. [X] autocomplete fields in the Contributor lookup are correctly displayed in UI
            2. [X] fields in the edit form are disabled
         2. **assert backend:**
            1. [X] Contributor **is NOT** created
            2. [X] ReleaseContributor **IS** created
            3. [X] update/edit is not allowed
      2. **Contributor is NOT USED in any release**
         1. **assert frontend:**
            1. [X] autocomplete fields in the Contributor lookup are correctly displayed in UI
         2. **assert backend:**
            1. [X] Contributor **is NOT** created
            2. [X] ReleaseContributor **IS** created
            3. [X] update/edit is allowed
   3. From **existing User:**
      1. **a Contributor with corresponding user_id EXISTS:**
         1. **assert frontend:**
            1. [X] fields are correctly prefilled
            2. [X] all UI fields are disabled
         2. **assert backend:**
            1. [X] Contributor **is NOT** created (existing Contributor is reused)
            2. [X] ReleaseContributor **IS** created
            3. [X] update/edit is not allowed
      2. **Contributor with corresponding user_id DOESN'T EXIST:**
         1. **assert frontend:**
            1. [X] fields are correctly prefilled
            2. [X] all UI fields are disabled
         2. **assert backend:**
            1. [X] Contributor **IS** created
            2. [X] ReleaseContributor **IS** created
            3. [X] update/edit is not allowed
2. CREATE DUPLICATE (by username, email...) CONTRIBUTORS is not allowed
   1. **assert frontend:**
      1. [X] show error message
   2. **assert backend:**
      1. [X] do not allow duplicates
3. EDIT CONTRIBUTOR 
   1. [X] Contributor updates/edits are allowed ONLY if Contributor belongs **ONLY TO ONE RELEASE** (this release)
4. DELETE CONTRIBUTOR 
   1. [X] check that Contributors can be deleted
   2. [X] check that the last Contributor cannot be deleted
5. REARRANGE CONTRIBUTORS 
   1. [X] check if drag and drop rearranging of Contributors in the UI is persisted in form of ReleaseContributor.index in the backend


django/library/serializers.py
```diff
@@ -110,19 +190,46 @@ def get_profile_url(self, instance):
             )
 
     def update(self, instance, validated_data):
-        affiliations = validated_data.pop("affiliations", None)
-        validated_data.pop("given_name", None)
-        validated_data.pop("family_name", None)
+        if not self._is_not_used_or_used_by_current_release_only(instance):
+            if self._trying_to_modify_attributes(instance, validated_data):
```
sgfost, line 194: does this update() get called when we do stuff like re-arranging/changing roles? I'm not sure what the `is_trying_to_modify..` check is for

sgfost (commented):
should we just rename `json_affiliations` to `affiliations`  now and change the old `affiliations` to `affiliation_tags` and add a deprecation comment/help_text? A migration needs to be added anyways

django/library/serializers.py
```diff
@@ -110,19 +190,46 @@ def get_profile_url(self, instance):
...
+        # if a contributor is made from a user, do not allow to modify it's data.
+        # POP following:
+        # 'given_name'
+        # 'middle_name'
+        # 'family_name'
+        # 'email'
+        # 'type'
+        # 'affiliations'
+        # 'json_affiliations'
+        user_id  = validated_data.pop("user_id", None)
```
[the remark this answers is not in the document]
asuworks: Since this is from the `update()` action and this only deals with the case of updating a Contributor which was created from an existing user, I think we shouldn't allow changing Contributor-AuthUser relation (user_id) in this case, and also any of the other user related attributes for the sake of consistency of Contributor-AuthUser fields. .
When a new Contributor is created from an existing AuthUser the user_id will be set inside the `create()` action which gets the whole `user` as parameter.

django/library/serializers.py
asuworks: yeah, `update()` is called when contributors are rearranged and roles are changed too. 
If a contributor is already "used" for another codebase, we still need to be able to add him to any other codebase, just need to make sure that the PUT request (which is called on that newly added contributor) doesn't modify any of it's attributes. Hence the naming: `_trying_to_modify_attributes()`

we are trying to squeeze to much functionality through one route `/contributors`...
Next improvement could be to split CREATE and UPDATE actions into 2 different routes in DRF.

django/library/serializers.py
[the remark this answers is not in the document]
sgfost: Does it remove the existing user relation from the contributor? It reads like that but idk if that actually happens

django/library/serializers.py
sgfost: wonder if its possible to change something in the ReleaseContributorSerializer so that `update()` on the contributor serializer doesn't get invoked when we are just changing the index/roles/etc

This probably works, the method to check whether we are updating in the update() method just felt wrong lol

django/library/serializers.py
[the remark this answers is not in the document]
asuworks: no, it doesn't remove the user_id from contributor

django/library/models.py
```diff
@@ -214,7 +234,7 @@ def to_codemeta(self):
         if self.orcid_url:
             codemeta["@id"] = self.orcid_url
         if self.affiliations.exists():
-            codemeta["affiliation"] = self.formatted_affiliations
+            codemeta["affiliation"] = self.formatted_json_affiliations
```
alee, line 237: this should be a rich object [schema.org/Organization](https://schema.org/Organization), maybe something like:

```
affiliation = self.json_affiliations[0] # assuming it exists
{
"@type": "Organization",
"@id": affiliation.ror_id,
"identifier": affiliation.ror_id,
"name": affiliation.name,
"url": affiliation.url,
"sameAs": affiliation.ror_id
}
```

alee (commented):
overall LGTM, thanks @asuworks !

after this gets merged and populate_contributor_affiliations is run we should set up a data migration to remove the old contributor.affiliations tags + models and rename json_affiliations -> affiliations as @sgfost mentioned but that can be in a followup PR + issue

alee:
a few minor notes:

1. adding a new contributor while searching for an existing User should use their affiliations to populate the affiliations search box if they already have them (either in MemberProfile or Contributor) but this kind of brings up some messy data duplication. How should we handle Contributor -> User -> `MemberProfile.affiliations` vs `Contributor.json_affiliations`? 
2. (minor issue) saving seems to take a while on staging but this could be unique to staging
3. in local testing using the "Search for an Existing User" enters details but disables all data entry fields which makes "adjustments" to the fields impossible and also makes the "Contributors can only be updated when exclusive to one codebase" error message unnecessary

image

4. on staging it appears that "Search for an Existing User" still allows for updates to those fields, not sure why there's a difference between local and staging?

image

sgfost:
An affiliations model either expanding the json (don't remember why it did get turned into a json field, now that I think about it) or just wrapping the json field is probably the only way to reasonably eliminate duplication there. I don't think theres any need to worry about syncing things aside from copying over `user.mp.affiliations` on contributor creation

I'd bet the deletion and re-creation of `ReleaseContributor`s for every request is making things a bit slower than they should be, one of the things to look into when refactoring the contributor route

alee:
duplicate set bug still present if you manage to click save twice (reproducible on staging where it takes a minute to process)

image

asuworks:
> a few minor notes:
> 
> 1. adding a new contributor while searching for an existing User should use their affiliations to populate the affiliations search box if they already have them (either in MemberProfile or Contributor) but this kind of brings up some messy data duplication. How should we handle Contributor -> User -> `MemberProfile.affiliations` vs `Contributor.json_affiliations`?

It could be the fact that the fake Contributors with a user_id from our DB have totally different values from the corresponding Users table. Could you show an example of duplication you are talking about?

> 2. (minor issue) saving seems to take a while on staging but this could be unique to staging
yes... 

> 3. in local testing using the "Search for an Existing User" enters details but disables all data entry fields which makes "adjustments" to the fields impossible and also makes the "Contributors can only be updated when exclusive to one codebase" error message unnecessary
> 

This is because of 'bad' data in our DB (you selected a user who already has a Contributor with that user_id, but their fields are totally different). Normally, this shouldn't happen in real life. So when you are trying to create add a Contributor from existing User following happens:
1. The form is populated from AuthUser (Contributor.json_affiliations will also be populated from MemberProfile.affiliations)
2. When you click "Save": a new Contributor and a new ReleaseContributor will be created with User data.

The idea behind disabling the UI fields is: we shouldn't allow changing fields if a Contributor is created from an existing User.

What we probably should do in PROD before deploying this, is to sync all fields for all existing Contributors with a user_id with the corresponding AuthUser fields + mp.affiliations. Copy fields from User to Contributor 

> 4. on staging it appears that "Search for an Existing User" still allows for updates to those fields, not sure why there's a difference between local and staging?

My bet is that James Martin has been created manually as a new Contributor and not through existing User. Can you try deleting him and adding him again. Are the fields still editable?

asuworks:
> duplicate set bug still present if you manage to click save twice (reproducible on staging where it takes a minute to process)
> 
> image

True story! This is a repeated request bug. I will make a quick fix in frontend (move "Save" button disabling right after it has been clicked)

What about the backend: should we use redis to store requests based on ip? Create some sort of middleware for repeated requests?

asuworks:
_is_not_used_or_used_by_current_release_only() was doing backend validation: is the user allowed to change the Contributor or not?

is_mutable() doesn't check whether the current user is allowed to modify the Contributor.

Example:
Say, User_A added to his CodebaseRelease_A a manually created Contributor_A (not used for any other Codebase). 
Say, User_B wants to add Contributor_A. `is_mutable()` will return True, since it is used in only one Codebase (Codebase_A) thus allowing to modify a foreign Contributor (which should not happen).

I do not see a way to avoid a DB call in is_mutable(), since we do not have the info (wether the incoming Contributor is used somewhere else or not).

Any other ideas?

sgfost:
I think the only other way of doing something similar without a query was to disallow updating at all once the contributor has been created (`obj.id is not None`). The case where a contributor needs to be updated because of a mistake of something is probably rare, but on the other hand will create junk data since they'll need to create another.

Only other idea I have (not new, this is or is similar to one of the first things @asuworks proposed) is to redo the contributor structure in order to simplify everything by 'flattening' release contributor and contributor into one model which has
- id
- release
- roles
- index
- include_in_citation
- affiliations
- user OR name and email (either don't store these for user-tied contributors or set it on create and make it immutable)

The downside to this was that it means 'duplicate' contributors for the same person, but all things considered now I think its much cleaner

alee:
> _is_not_used_or_used_by_current_release_only() was doing backend validation: is the user allowed to change the Contributor or not?
> 
> is_mutable() doesn't properly check whether the current user is allowed to modify the Contributor. Contributor.user_id could help us avoid a DB call (is_mutable will return False if user_id is not None) only in the case when Contributor was created from a User. What about other cases? (Manually created Contributor, contributor from existing Contributor without user_id?)
> 
> Example: Say, User_A added to his CodebaseRelease_A a manually created Contributor_A (not used for any other Codebase). Say, User_B wants to add Contributor_A. `is_mutable()` will return True, since it is used in only one Codebase (Codebase_A) thus allowing to modify a foreign Contributor (which should not happen).
> 
> I do not see a way to avoid a DB call in is_mutable(), since we do not have the info (wether the incoming Contributor is used somewhere else or not).
> 
> Any other ideas?

I don't think we need to care - the _only_ case where we might want to allow editing of an existing Contributor is for the person who originally submitted that Contributor. All other cases, make it immutable. But since we aren't capturing that information I think we should simplify it to, "All existing Contributors or Contributors created from a User should be immutable". The only other situation where it might make sense for mutability is if we are adding data, like Affiliations, but I think it's fine to simply leave it out.

@sgfost I think the flatten Contributors and ReleaseContributors path is a non-starter, we're already overdue on this particular task which was a bit of an iceberg tbf and I'm not entirely convinced that flatteing them is a cleaner design either, just favoring one set of design tradeoffs for another.

asuworks:
That will make things a lot easier, but less intuitive: why can't I edit a Contributor, I have just manually created?
In the frontend: "All existing Contributors or Contributors created from a User should be immutable" will translate to basically not allowing any edits at all.

```
def is_mutable():
if obj.id is not None:
    return False
```

Is that what we want?

asuworks:
in fact, we could skip the update() method completely

sgfost:
>All existing Contributors or Contributors created from a User should be immutable

I think that is perfectly fine for now. Just show a note on the form or only show roles/citation when editing

>in fact, we could drop the update() method completely

Yep, was just typing this out, `instance.id` should ever be `None` in `update()`. Ideally completely disable update by raising an error but since this get called regardless I think we still need that 'is modifying' check

```python
def update(self, instance, validated_data):
  if trying_to_modify(instance, validated_data):
    raise ValidationError(..) # or MethodNotAllowed?
  return instance
```

frontend/src/components/releaseEditor/ContributorEditForm.vue
```diff
@@ -111,7 +111,7 @@
     </form>
   </div>
   <div class="modal-footer border-0">
-    <button type="button" class="btn btn-outline-gray" @click="resetContributor">Reset</button>
+    <button v-if="!disableEditForm" type="button" class="btn btn-outline-gray" @click="resetContributor">Reset</button>
```
alee, line 114: IIRC this unfortunately makes the reset button disappear once you search for and select a user in the "Create New Contributor" pathway which the only use case you'd want for to begin with

PR #725, opened 2024-05-27
feat: add cypress end to end tests

CharlesSheelam:
Working on smoke testing to test functionality of the comses website.


e2e/cypress/fixtures/codebase/testNarrativeDocumentation.txt
```diff
@@ -0,0 +1 @@
+//test narative documentation
```
alee, line 1: narrative

e2e/cypress/fixtures/codebase/testSourceCode.txt
```diff
@@ -0,0 +1,2 @@
+//test source code
+
```
alee, line 2: Should we include some relatively short public domain source code here instead?

e2e/cypress/tests/beforeLoginTests.spec.ts
```diff
@@ -0,0 +1,19 @@
+//test ability to browse/download codebases
+describe("Browse Codebases", () => {
+  it("should browse the model library and be able to enter a search query", () => {
+    cy.visit("/"); // visit homepage
+    cy.visit("/codebases");
+    assert(cy.get("h1").contains("Computational Model Library"));
```
alee, line 6: should we also make assertions on an expected model to find in the model library?

e2e/cypress/tests/beforeLoginTests.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  });
+});
+
+//download codebase
+describe("Download Codebases", () => {
+  it("should download a codebase", () => {
+    // Navigate to the first 'search-result', find the 'a' tag within the 'title' div, and click it
+    cy.get(".search-result", { timeout: 10000 })
+      .first()
+      .find(".title a")
```
alee, line 16: this filtering criteria seems poorly vague for identifying a codebase download. Setup a more precise criteria even if you need to include additional metadata on the model library page.

e2e/cypress/tests/codebase.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  //codebases PAGE
+
+  it("should visit the codebases page", () => {
+    cy.visit("/codebases");
+    assert(cy.get("h1").contains("Computational Model Library"));
+  });
+
+  it("should  be able to download a codebase", () => {
+    cy.visit("/codebases");
+    cy.get(".search-result").first().find("a").first().click();
```
alee, line 22: How brittle is this criteria? Is the first a href after a `.search-result` CSS class targeted enough to always get the first codebase? Or should we make all codebase results more explicit with a CSS class or identifier and target that explicitly?

e2e/cypress/tests/codebase.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      .click();
+    cy.get('[data-cy="save-and-continue"]').click();
+    cy.get('button.btn.btn-danger[rel="nofollow"]').click();
+    cy.get('button[type="submit"].btn.btn-danger[form="publish-form"]').click();
+  });
+
+  it("should verify that the codebase was uploaded correctly", () => {
+    cy.visit("/codebases");
+    cy.get(".search-result").first().find("a").first().click();
+    assert(cy.get("h1").contains("Codebase Title"));
```
alee, line 114: These kinds of assertions are highly dependent on whatever stylistic or textual changes we want to make in the frontend. It might be better to have a list or other data structure of expected text and programmatically iterate through it with assertions and comment the data structure as a place of possible future change.

alee (changes requested):
Comments attached, what do you think @sgfost ?

e2e/cypress/tests/codebase.spec.ts
sgfost: Good idea, perhaps a global object that describes the initial database state which can be modified at runtime? for example:

```js
{
  ...
  codebases: [
    { title: "Codebase Title", description: "something" },
  ],
  users: [
    { username: "test_user", password: "123456", ... }
  ],
  ...
}
``` 

and then, as Allen said, use this to assert that certain things are found on a page, rather than at some specific location in the DOM tree

e2e/cypress/tests/codebase.spec.ts
sgfost: I think this and the above comment are the same issue, should probably just use data attributes to target this

e2e/cypress/tests/user.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+        cy.visit("/users");
+        cy.get("#djHideToolBarButton").click();
+        cy.contains('div', 'My profile').click();
+        cy.contains('a', 'Edit Profile').click();
+        cy.get('#form-field-givenName').clear()
+        cy.get('[data-cy="first name"]').type("New first name");
+        cy.get('#form-field-familyName').clear()
+        cy.get('[data-cy="last name"]').type("New last name");
+        cy.get('[data-cy="submit"]').click();
+        assert(cy.get("h4").contains("New first name New last name"));
```
sgfost, line 17: another example of somewhere we'd want to just do a lookup on the page for the text _somewhere_ rather than in this specific tag

e2e/cypress/tests/codebase.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      .get("button")
+      .first()
+      .click({ force: true });
+    cy.get('button.btn.btn-outline-gray[data-bs-dismiss="modal"]')
+      .should("be.visible")
+      .and("not.be.disabled")
+      .first()
+      .click({ force: true });
+    cy.get("body").click(0, 0);
+    cy.get("body").click(0, 0); //FIX: find a more precise way of closing the image upload modal
```
sgfost, line 68: I think this is alright, its the behavior we want from a modal

sgfost (changes requested):
thanks @CharlesSheelam. I think the main thing remaining is fragility in selectors.

For cases where we need to select an element to perform some action (like grabbing a codebase search result and the link to it's detail page) it should look for a data attribute rather than a css class/first element/etc. 

And for cases where we're asserting that something exists, I would say just expand the search to be much more broad (whole page or perhaps an existing container with a data attribute) and use a global data structure that describes the content of the database to define what we want to check for

sgfost:
tests are failing when running on the GH runner, partially because I disabled the debug toolbar but there are a few other things it seems to be getting tripped up on. They mostly seem to coincide with selectors mentioned above that should get changed anyways

alee (approved):
LGTM, only issue is the data-cy attribute values are sometimes space cased or kebab-cased, should stick to one convention. Can fix in a future refactor

sgfost:
failing e2e tests is due to programming error on my part in #757. I suppose it illustrates the usefulness of these!

sgfost (approved):

alee (approved):
thanks @sgfost and @CharlesSheelam !

PR #735, opened 2024-07-10
build: Change Vite server port to 5173

Karthik99999:
Port 5000 seems to be reserved by some MacOS process (AirTunes is referenced in the response headers). Changing the port should allow the Vite server to successfully be accessed by the django server on MacOS.

Resolves comses/infra#192


sgfost:
thx!

alee (approved):
good work sleuthing that one out, it seemed like the vite server was binding just fine to 5000 from the logs but looking at https://stackoverflow.com/questions/69818376/localhost5000-unavailable-in-macos-v12-monterey and https://stackoverflow.com/questions/72369320/why-always-something-is-running-at-port-5000-on-my-mac + other posts it's definitely a thing!

alee:
only nitpick I'd add to this for the future is to consider making this an environment variable in dev.yml and thread it through to the server and vite containers so we can change it in one place if it ever happens again

PR #743, opened 2024-07-17
fix: use csv.writer instead of COPY TO in export_raw_data

sgfost:
perhaps I was just doing something wrong but the `COPY TO` in `export_raw_data` mgmt command didn't seem to work anymore (filesystem access issue?)

### DIFF
--- a/django/curator/management/commands/export_raw_data.py
+++ b/django/curator/management/commands/export_raw_data.py
@@ -1,3 +1,4 @@
+import csv
 import logging
 import os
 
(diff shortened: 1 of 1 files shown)


alee (commented):
COPY is a server-side command so the /shared directory needs to get mounted in the db. Might revert this and add it to docker-compose configuration instead since it's much more efficient to run it on the db directly

sgfost:
fixed by 32ea37d

PR #746, opened 2024-07-25
footer rework

sgfost:
Re-design of the footer to have a slightly more manageable section for funding/support logos. Includes some small tweaks to certain logos for the sake of coherence

Also adds shadows to banners with image backgrounds (intros, featured models, footer) for some depth but this is more subjective

desktop:

image

mobile:

image


alee:
Looks a little weird on my browser:

image

sgfost:
sure does.. Though it looks fine on my chrome. Cached stylesheets maybe? That's what it looks like if the changes to _home.scss are reverted

sgfost:
Though vite should be cache busting, not sure how that would happen

alee:
looks fine now after a ctrl+shift+r hard refresh

alee (approved):
thanks Scott! need to upgrade our social icons but that might go along with an upgrade to bootstrap 5 or major change migration to tailwind

cmbarton:
Looks nice. We could the current reference to the E. Musk service seen in most news stories. "X (formerly known as Twitter)"

PR #748, opened 2024-07-31
feat: Add PeerReviewer model for improved peer review support

Karthik99999:
see comses/planning#256

This PR currently does not include the functionality of users being able to request to become peer reviewers.


frontend/src/components/ReviewersPage.vue
```diff
@@ -0,0 +1,81 @@
...
+              editFormRef?.resetForm();
+              editModal?.show();
+            "
+            >Edit</a
+          >
+          <a class="btn btn-danger" @click="deleteReviewer(reviewer)">Delete</a>
+        </div>
+      </div>
+    </div>
+    <div class="col-sm-12 col-md-3">
```
sgfost, line 29: the filtering can be here in this column

frontend/src/components/ReviewersPage.vue
```diff
@@ -0,0 +1,81 @@
...
+          <a
+            class="btn btn-primary"
+            @click="
+              editCandidate = reviewer;
+              editFormRef?.resetForm();
+              editModal?.show();
+            "
+            >Edit</a
+          >
+          <a class="btn btn-danger" @click="deleteReviewer(reviewer)">Delete</a>
```
sgfost, line 25: for inactive reviews this can do the opposite

django/library/views.py
```diff
@@ -171,14 +173,41 @@ def post(self, request, *args, **kwargs):
...
+        query = request.query_params.get("query", None)
+        if query is None:
+            return queryset
+        return get_search_queryset(query, queryset)
+
+
+class PeerReviewerViewSet(CommonViewSetMixin, NoDeleteViewSet):
+    queryset = PeerReviewer.objects.exclude(is_active=False)
+    serializer_class = PeerReviewerSerializer
+    permission_classes = (ChangePeerReviewPermission,)
```
sgfost, line 206: We'll need to allow a user to create/update a peerreviewer record for themselves. I think the way to do this would be overriding `perform_create()` and `perform_update()` here to do the permission checks rather than using the permission class

sgfost (changes requested):
Thanks Karthik, all of the basic functionality seems to be working. See the inline comments for remaining stuff

frontend/src/components/ReviewersPage.vue
Karthik99999: So should inactive reviewers still show up in the list (maybe with (Inactive) added beside their name)?

sgfost: should be some indicator in the list but also filtered out by default with the option to un-select that filter

frontend/src/components/ReviewersPage.vue
```diff
@@ -0,0 +1,81 @@
...
+        </div>
+      </div>
+    </div>
+    <div class="col-sm-12 col-md-3">
+      <ReviewerAddModal />
+    </div>
+    <BootstrapModal
```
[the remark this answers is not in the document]
Karthik99999: I just removed the extra component, makes it easier to call `retrieveReviewers`

django/library/views.py
```diff
@@ -171,14 +173,41 @@ def post(self, request, *args, **kwargs):
...
+    def filter_queryset(self, request, queryset, view):
+        # FIXME: when there query, return the full list of reviewers
+        query = request.query_params.get("query", None)
+        if query is None:
+            return queryset
+        return get_search_queryset(query, queryset)
+
+
+class PeerReviewerViewSet(CommonViewSetMixin, NoDeleteViewSet):
+    queryset = PeerReviewer.objects.exclude(is_active=False)
```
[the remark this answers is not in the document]
Karthik99999: How would the filtering work here? The main reviews dashboard uses a form but I think that would require using jinja templates.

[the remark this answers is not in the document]
sgfost: https://www.comses.net/codebases/ has a UI for filtering done in vue (`CodebaseListSidebar.vue`) which just adds on query params to the search/list url which get picked up by the `CodebaseFilter`. Could do the same thing here OR could do the filtering all in client-side JS so its faster

As long as its relatively simple to implement, I think doing it all client-side makes more sense since its a short list and non-paginated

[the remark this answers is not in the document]
sgfost: the filters I would do for now are:

- include inactive (checkbox)
- name (search box, partial match on name/username)
- language (checkboxes, based on the set of languages available in the whole list)

it might be too messy to try and re-use the existing field components for this, btw

sgfost:
not sure whats going on with the tests action here. It looks like the unrelated issue comses/planning#233 but that seemed to only happen rarely. Don't see anything in the changes here that would mess with anything

django/library/jinja2/library/review/reviewers.jinja
```diff
@@ -0,0 +1,13 @@
+{% extends 'base.jinja' %}
+
+{% block introduction %}
+    <h1>Manage Peer Reviewers</h1>
+{% endblock %}
+
+{% block content %}
```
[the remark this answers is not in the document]
sgfost: maybe tabs (https://getbootstrap.com/docs/5.2/components/navs-tabs/#tabs) at the top like with other pages

frontend/src/components/ProfileEditReviewer.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <ReviewerCard
+    v-if="reviewer"
```
sgfost, line 3: there should be a better indication that the reviewer record is inactive here. Currently it just shows the "activate" button. Maybe a badge for `active`/`inactive` on the reviewer card card-title in general and possibly hiding everything besides a `Reactivate` button here on the profile

frontend/src/components/ReviewerCard.vue
```diff
@@ -0,0 +1,44 @@
...
+        <strong>Email:</strong> {{ reviewer.memberProfile.email }}<br />
+        <strong>Programming Languages:</strong> {{ reviewer.programmingLanguages.join(", ") }}<br />
+        <strong>Subject Areas:</strong> {{ reviewer.subjectAreas.join(", ") }}<br />
+        {{ reviewer.notes }}
+      </p>
+      <a class="btn btn-primary me-1" @click="emit('edit')">Edit</a>
+      <a v-if="reviewer.isActive" class="btn btn-danger" @click="changeActiveState(reviewer, false)"
+        >Deactivate</a
+      >
+      <a v-else class="btn btn-success" @click="changeActiveState(reviewer, true)">Activate</a>
```
sgfost, line 19: minor aesthetic gripe: the two different greens side by side look off to me and the buttons are a bit large. How about the following, which uses the same styling as the contributor cards in the release editor:

```html
<template>
  <div class="card mb-3">
    <div class="card-header d-flex justify-content-between align-items-center">
      <h5 class="card-title mb-0">
        {{ reviewer.memberProfile.name }} ({{ reviewer.memberProfile.username }})
      </h5>
      <span>
        <a class="btn btn-link" @click="emit('edit')">
          <small><i class="fas fa-edit me-1"></i>Edit</small>
        </a>
        <a
          v-if="reviewer.isActive"
          class="btn btn-link text-danger"
          @click="changeActiveState(reviewer, false)"
        >
          <small><i class="fas fa-trash me-1"></i>Deactivate</small>
        </a>
        <a v-else class="btn btn-link text-success" @click="changeActiveState(reviewer, true)">
          <small><i class="fas fa-check-circle me-1"></i>Activate</small>
        </a>
      </span>
    </div>
    <div class="card-body">
      <p class="card-text">
        <strong>Email:</strong> {{ reviewer.memberProfile.email }}<br />
        <strong>Programming Languages:</strong> {{ reviewer.programmingLanguages.join(", ") }}<br />
        <strong>Subject Areas:</strong> {{ reviewer.subjectAreas.join(", ") }}<br />
        {{ reviewer.notes }}
      </p>
    </div>
  </div>
</template>
```

sgfost (changes requested):
Nice job, everything seems to work well, besides a small issue with the reviewers data migration. I have a comment above about this along with a few other things

frontend/src/components/ReviewerCard.vue
sgfost: image

frontend/src/components/ProfileEditReviewer.vue
Karthik99999: image
Does something like this work?

sgfost: yeah that works, good call.

Since the edit is hidden we can use the regular blue link color, and maybe change the icon + wording

```html
        <a v-else class="btn btn-link" @click="changeActiveState(reviewer, true)">
          <small><i class="fas fa-sync-alt me-1"></i>Reactivate</small>
        </a>
```

sgfost (approved):

django/library/models.py
```diff
@@ -2001,7 +2002,10 @@ def get_change_closed_url(self):
         return reverse("library:peer-review-change-closed", kwargs={"slug": self.slug})
 
     def get_invite(self, member_profile):
-        return self.invitation_set.filter(candidate_reviewer=member_profile).first()
+        # return self.invitation_set.filter(candidate_reviewer=member_profile).first()
+        return self.invitation_set.filter(
```
alee, line 2006: should this be ordered by most recent e.g., `self.invitation_set.order_by('-date_created').filter(...).first()`

could also add a default `ordering` to `PeerReviewInvation` class Meta

frontend/src/components/form/TextListField.vue
```diff
@@ -66,14 +66,8 @@ export interface TextListFieldProps {
...
 
-onMounted(() => {
-  if (!value.value) {
-    // force initialize to empty array
-    value.value = [];
-  }
-});
-
 function create() {
+  if (!value.value) value.value = [];
```
alee, line 70: `value.value` would benefit from improved naming

alee (approved):
Thanks Karthik & Scott! LGTM with some minor comments

frontend/src/components/form/TextListField.vue
sgfost: `ref()` weirdness, I'll think of something.. I don't know if I want to start calling every var `x: Ref` `xRef: Ref`. Maybe just `fieldValue` in this case

sgfost:
I'll fix these up to get it moving. There's also a small oversight in the browser tests with dates which is causing that to fail today

alee:
thanks @sgfost ! I was wondering about the e2e tests referring to dates with scalar integers, https://github.com/comses/comses.net/blob/ff5c42d60824f81034788fedd46ea45b8f869b64/e2e/cypress/fixtures/data.json#L45

sgfost:
the least fragile way I can think of doing it (i.e. not relying on this specific datepicker too much) is to keep the scalar day-of-the-month input but always using the following month. It'd probably be ideal if there was a way to input a string date with the keyboard but I'll just fix the test for now rather than hack at the datepicker

PR #749, opened 2024-08-02
improve search

asuworks:
trying to close https://github.com/comses/planning/issues/21

### DIFF
--- a/django/core/view_helpers.py
+++ b/django/core/view_helpers.py
@@ -3,26 +3,73 @@ from django.core.exceptions import FieldError
 from rest_framework.response import Response
 from wagtail.search.backends import get_search_backend
 from wagtail.contrib.search_promotions.models import Query
-from wagtail.search.query import MATCH_ALL, Phrase
-from wagtail.search.utils import parse_query_string
+from wagtail.search.query import (
+    MATCH_ALL,
+    Phrase,
+    Boost,
+    Or,
+    Fuzzy,
+    SearchQuery,
+)
+
+from itertools import combinations
 
 from .models import ComsesGroups
 
 logger = logging.getLogger(__name__)
 
-search_backend = get_search_backend()
+
+def build_search_query(input_text: str) -> SearchQuery:
+    words = input_text.split()
+
+    # Highest priority: Fuzzy match of the entire input text
+    fuzzy_full_query = Boost(Fuzzy(input_text), boost=20.0)
+
+    # Second priority: Exact match using Phrase
+    exact_match_query = Boost(Phrase(input_text), boost=15.0)
+
+    # Generate all 2-word combinations
+    two_word_combos = list(combinations(words, 2))
+
+    # Create queries for 2-word combinations
+    two_word_queries = []
+    for combo in two_word_combos:
+        two_word_phrase = " ".join(combo)
+        # Boost 2-word phrases higher if both words are in the original text
+        if all(word in input_text for word in combo):
+            two_word_queries.append(Boost(Phrase(two_word_phrase), boost=12.0))
+        else:
+            two_word_queries.append(Boost(Phrase(two_word_phrase), boost=3.0))
+
+    # Make the search fuzzy using Fuzzy for individual words if word length > 3, otherwise use Phrase
+    single_word_queries = [
+        Fuzzy(word) if len(word) > 3 else Phrase(word) for word in words
+    ]
+
+    # Combine all queries
+    all_queries = (
+        [fuzzy_full_query, exact_match_query] + two_word_queries + single_word_queries
+    )
+
+    # Combine all queries using Or
+    combined_query = Or(all_queries)
+
+    return combined_query
 
 
 def get_search_queryset(
     query, queryset, operator="or", fields=None, tags=None, criteria=None
 ):
+    search_backend = get_search_backend()
+
+    if not fields:
+        fields = []
+
     if not tags:
         tags = []
 
     if not criteria:
         criteria = {}
-
-    order_by_relevance = not queryset.ordered
 
     """
     # FIXME: this won't work until RelatedFields support filtering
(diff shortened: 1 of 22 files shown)


django/library/jinja2/library/codebases/list.jinja
```diff
@@ -45,5 +45,12 @@
 {% endblock %}
 
 {% block js %}
+<script nonce="{{ request.csp_nonce }}" >
+    {% if language_facets %}
```
sgfost, line 49: can this be passed in as a data attribute on the sidebar div instead? theres a function `extractDataParams` to help with this on the js side

asuworks: you mean without attaching language_facets to window global variable?

sgfost: yeah like

```html
<div id="sidebar" {% if language_facets %} data-language-facets="{{ language_facets }}" {% endif %} />
```

mostly for consistency, not sure if there's any meaningful practical difference

asuworks: In what file would you do this? As I understand this, we'd need to do this in jinja, but we need the data in the vue component... I am lost here.. help, please!

sgfost: this one, the js side parses it with `extractDataParams` before mounting the component and passing the extracted stuff as props. Most mounting scripts in `/apps` do it

asuworks: oh, got it! thanks!

django/core/view_helpers.py
```diff
@@ -33,24 +78,48 @@ def get_search_queryset(
...
+    Filter queryset 
+    """
     if criteria:
+        logger.debug(f"criteria={criteria}")
         try:
             queryset = queryset.filter(**criteria)
-        except FieldError:
-            logger.warning("Invalid filter criteria: %s", criteria)
+        except FieldError as e:
+            logger.warning(f"Invalid filter criteria: {criteria}. Error: {str(e)}")
```
alee, line 109: use `exc_info=e` instead of converting the exception into a string and missing out on the entire stack trace

frontend/src/composables/api/codebase.ts
```diff
@@ -3,10 +3,12 @@ import { useAxios, type RequestOptions } from "@/composables/api";
 
 interface CodebaseQueryParams {
   query?: string;
-  publishedAfter?: string;
-  publishedBefore?: string;
+  published_after?: string;
```
alee, line 6: is this to line up with the incoming query parameters?

alee (changes requested):
thanks @asuworks this looks great! still need to test & deploy locally but some minor coding style recommendations.

I think we should try to keep our logging formats consistent throughout the codebase but the format strings are definitely easier to read with lots of parameters...

sgfost (commented):
Nice. Filter sidebar looks great and the search querying anecdotally seems a lot better

The default query should probably go back to ordering things by publish date descending though

I like the new selected filter terms display better than the one that comes from the paginator macro at the top. In the case of codebases should we disable those and just use the sidebar ones so we aren't showing that information in two places? If so it could:
1. get moved to the top of the sidebar
2. only show up when those filters are applied
3. have a button to clear all

asuworks:
> thanks @asuworks this looks great! still need to test & deploy locally but some minor coding style recommendations.
> 
> I think we should try to keep our logging formats consistent throughout the codebase but the format strings are definitely easier to read with lots of parameters...

thanks @alee! fixed it!

asuworks:
@sgfost 

1. agree about the default ordering by Publish date: newest 
2. agree about removing the filter tags and the red clear button from under the searchbar. Leave only the text: `Displaying 4 of 4 results for 'lala'` @alee ?
3. moving selected filters in the sidebar to the top would make the filtering ui elements jump up and down.. I'd say it's better to keep the selected filter tags at the bottom.

sgfost:
the issue I have is that the whole sidebar is like 1000px tall without filters so you have to scroll up and down to do stuff, maybe it can be shrunk down to something more manageable? If the selected filters only appear when applied (which I think is what should happen) it may be less of an issue since the page would reload anyways.

Could also just apply on selection, I don't think we have nearly enough different filters to make that a problem, even without AJAX searching

asuworks:
how about programming languages in 2 columns?

image

asuworks:
we could also remove the "Publish a model" button (I think it is not very necessary there)

sgfost:
the publish button should be somewhere on that page, maybe gets moved to whatever we replace that blue alert box with. The counts wrapping over looks a bit odd, what about collapsing it by default? Or showing just a few with a `show more` button.

Other stuff could probably be shrunk down too, padding, margins, text size, etc. The publish date filters can be together in one row

asuworks:
image

asuworks:
much nicer and shorter :)

asuworks:
the wrapping is not nice, agree... maybe back to single column? The collapsing of the programming languages would complicate the ui I think...
image

sgfost:
pretty sure thats the only link to publishing a model, I wouldn't want to suddenly move it. Regardless it still hardly fits on my screen

How does this UI look? I squished things down a bit and removed the apply button assuming we can apply everything on selection. If not, then the selected filters can sit at the bottom with the apply button since most screens should have the full thing in view. There should at least be a way to clear them from the top though

Screenshot 2024-08-05 at 17-02-34 CoMSES Net Computational Model Library

asuworks:
@sgfost I implemented your version. When selecting a checkbox or a radio in the sidebar, the ui jumps down because the selected filters are populated at the top... maybe put them back at the bottom?

sgfost:
I had in mind only displaying them when they are actually applied. I suppose that may still pose an issue if things are applied on selection.. Its probably fine at the bottom as long as the whole thing isn't too tall

asuworks:
Today we talked about showing selected filter badges in the sidebar on page load...
I decided to get rid of them completely, since it seemed like an unnecessary duplication of the badges we display under the searchbar.

alee:
@all-contributors please add @sgfost for doc, review, maintenance, mentoring

alee:
@all-contributors please add @asuworks for code, doc

PR #754, opened 2024-08-08
Add mark spam button

ldwiyono:
Adds mark_spam_confirm_modal in core folder to be implemented for Events, Jobs, Codebases. Essentially resolves comses/planning/issues/255, however improvements in UX and further functionalities are still possible.

Current appearance on desktop for events:
image

### DIFF
--- a/django/core/jinja2/core/events/retrieve.jinja
+++ b/django/core/jinja2/core/events/retrieve.jinja
@@ -75,7 +75,7 @@
         <a href="{{ url('core:event-edit', pk=id) }}">
             <div class="btn btn-primary my-1 w-100">Edit</div>
         </a>
-        {% if (user.is_superuser or user.groups.filter(name='Moderators').exists()) and not is_marked_spam %}
+        {% if can_moderate and not is_marked_spam %}
             {{ mark_spam_confirm_modal("Mark spam", title, url("core:event-mark-spam", pk=id), csrf_input) }}
         {% endif %}
     {% endif %}
--- a/django/core/jinja2/core/events/retrieve.jinja
+++ b/django/core/jinja2/core/events/retrieve.jinja
@@ -1,5 +1,5 @@
 {% extends "sidebar_layout.jinja" %}
-{% from "common.jinja" import breadcrumb, embed_discourse_comments, share_card, member_profile_href, search_tag_href, delete_confirm_modal, render_ogp_tags, alert_if_spam, alert_if_deleted %}
+{% from "common.jinja" import breadcrumb, embed_discourse_comments, share_card, member_profile_href, search_tag_href, delete_confirm_modal, render_ogp_tags, alert_if_spam, alert_if_deleted, mark_spam_confirm_modal %}
 
 {% block title %}{{ title }}{% endblock %}
 
@@ -75,6 +75,9 @@
         <a href="{{ url('core:event-edit', pk=id) }}">
             <div class="btn btn-primary my-1 w-100">Edit</div>
         </a>
+        {% if user.is_superuser and not is_marked_spam %}
+            {{ mark_spam_confirm_modal("Mark spam", title, url("core:event-mark-spam", pk=id), csrf_input) }}
+        {% endif %}
     {% endif %}
     {% if has_delete_perm %}
         {{ delete_confirm_modal("Delete Event", title, url("core:event-delete", pk=id), csrf_input) }}
--- a/django/core/jinja2/core/events/retrieve.jinja
+++ b/django/core/jinja2/core/events/retrieve.jinja
@@ -75,12 +75,12 @@
         <a href="{{ url('core:event-edit', pk=id) }}">
             <div class="btn btn-primary my-1 w-100">Edit</div>
         </a>
-        {% if can_moderate and not is_marked_spam %}
-            {{ mark_spam_confirm_modal("Mark spam", title, url("core:event-mark-spam", pk=id), csrf_input) }}
-        {% endif %}
     {% endif %}
     {% if has_delete_perm %}
         {{ delete_confirm_modal("Delete Event", title, url("core:event-delete", pk=id), csrf_input) }}
+    {% endif %}
+    {% if can_moderate and not is_marked_spam %}
+            {{ mark_spam_confirm_modal("Mark spam", title, url("core:event-mark-spam", pk=id), csrf_input) }}
     {% endif %}
     {#{{ share_card(absolute_url) }}#}
 {% endblock %}
--- a/django/core/jinja2/core/events/retrieve.jinja
+++ b/django/core/jinja2/core/events/retrieve.jinja
@@ -75,7 +75,7 @@
         <a href="{{ url('core:event-edit', pk=id) }}">
             <div class="btn btn-primary my-1 w-100">Edit</div>
         </a>
-        {% if user.is_superuser and not is_marked_spam %}
+        {% if (user.is_superuser or user.groups.filter(name='Moderators').exists()) and not is_marked_spam %}
             {{ mark_spam_confirm_modal("Mark spam", title, url("core:event-mark-spam", pk=id), csrf_input) }}
         {% endif %}
     {% endif %}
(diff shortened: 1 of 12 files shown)


django/core/mixins.py
```diff
@@ -268,6 +270,32 @@ def _validate_content_object(self, instance):
                     f"instance {instance} does not have a {field} attribute"
                 )
 
+    @action(detail=True, methods=["post"])
+    def mark_spam(self, request, **kwargs):
+        # consider adding notes field to the request
```
sgfost, line 275: can remove

django/core/mixins.py
```diff
@@ -268,6 +270,32 @@ def _validate_content_object(self, instance):
...
+                "notes": notes,
+            },
+        )
+        if not created:
+            spam_moderation.status = SpamModeration.Status.SPAM
+            spam_moderation.detection_method = "manual"
+            spam_moderation.reviewer = request.user
+            spam_moderation.notes = notes
+            spam_moderation.save()
+        # return Response({"message": "Successfully marked as spam!"}, status=status.HTTP_200_OK)
```
sgfost, line 296: this too

django/core/mixins.py
```diff
@@ -268,6 +270,32 @@ def _validate_content_object(self, instance):
                     f"instance {instance} does not have a {field} attribute"
                 )
 
+    @action(detail=True, methods=["post"])
```
sgfost, line 273: the action decorator can take in `permission_classes` to override the viewset level permissions https://www.django-rest-framework.org/api-guide/viewsets/

django/core/jinja2/core/events/retrieve.jinja
```diff
@@ -75,6 +75,7 @@
         <a href="{{ url('core:event-edit', pk=id) }}">
             <div class="btn btn-primary my-1 w-100">Edit</div>
         </a>
+        {{ mark_spam_confirm_modal("Mark spam", title, url("core:event-mark-spam", pk=id), csrf_input) }}
```
sgfost, line 78: should update the conditional rendering of the mark spam buttons to only show when the content object `is_marked_spam` is `False` and that the viewing user has permission

sgfost (changes requested):
I just realized there are no permissions required to use the `mark_spam` action so anyone can mark anything as spam (though the button is only shown to users with edit perms for the object, so admins and submitters)

The action should be locked down and the button should only be rendered when the user has the same permission. Not 100% sure what that should be, however.

@alee What do you think the access here should be: admins, wagtail moderators, editors, or a granular permission?

Otherwise it looks good, just a couple comments to get rid of

sgfost:
I think this would resolve comses/planning#201 which appears to talk about the `Moderators` group

ldwiyono:
I'm ready to make these changes whenever necessary.  The conditional could be something like `{% if has_change_perms and has_moderator_access %}`?

sgfost:
change perms aren't necessary. The condition on the template should be something like `IF NOT marked spam AND is a moderator`

I'd just go with the Moderator group for now and we can change it if needed. The permission class to assign to `mark_spam` should be straightforward, just need to make a custom DRF permission class (https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions) that checks if the `request.user` is a Moderator https://stackoverflow.com/a/20110261

ldwiyono:
Hi, I added the new permission class `MarkSpamPermissions` and locked down both the `mark_spam` action and the conditional rendering of the `mark_spam` button. I used the `is_superuser` attribute of the user for the permissions class; if this differs from "moderator" in significant ways, please let me know as I could not easily find a test for that group.

sgfost:
`is_superuser` is a field on the django user that makes `has_perm()` always return true, the idea being that it's one thing to set to give a user the ability to do anything on the site. The `Moderator` role refers to a `Group` which is provided by wagtail and has slightly less access than a wagtail admin.

You'll just need to check whether the user is a part of the `"Moderators"` group https://stackoverflow.com/questions/4789021/in-django-how-do-i-check-if-a-user-is-in-a-certain-group. Should probably add moderators to the `ComsesGroups` enum if we are going to start using it explicitly for stuff

sgfost:
also, no need to keep amending and force pushing the same commit. The initial rebase was just to take care of the test commits

ldwiyono:
Sorry about force pushing the commits, I'll keep that in mind for the future. I'm currently trying this:

```
class MarkSpamPermissions(permissions.BasePermission):
    def has_permission(self, request, view):
        if not request.user or not request.user.is_authenticated:
            return False
        
        return request.user.groups.filter(name='Moderator').exists()

```
but when I'm logged in as admin_user, it denies me permission to call the ```mark_spam``` action.
image

sgfost:
the `Admins` group (also from wagtail) is different from the `is_superuser` and `is_staff` flags on the user object. The admin_user must not have any of these set is and relying on being a superuser for access to everything. I think the solution would be to return true if `request.user.is_superuser` in the permission class before checking for group membership

ldwiyono:
Okay, thanks. I've incorporated those changes into the newest commit; please let me know if this works for you.

sgfost (approved):

alee (changes requested):
good work, thanks @ldwiyono and @sgfost !

some minor suggestions included inline, I think things would be cleaner if we pushed the `user.groups.filter(name=...).exists()` logic into `ComsesGroups.is_moderator(user)`

I've deployed for further testing on staging before merging. DM me on slack if you need access to a Moderator user account there, create a new user account if you want and I can add you to the group.

One thing immediately of note: I don't seem to be able to mark older events as spam (but ironically I can mark my own submissions as spam)

image

django/core/view_helpers.py
```diff
@@ -79,4 +79,8 @@ def add_change_delete_perms(instance, data, user):
         "{}.delete_{}".format(instance._meta.app_label, instance._meta.model_name),
         instance,
     )
+    data["can_moderate"] = (
+        user.is_superuser
+        or user.groups.filter(name=ComsesGroups.MODERATOR.value).exists()
```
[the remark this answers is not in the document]
sgfost: oh right, should we just replace and migrate the Reviewer group in #748?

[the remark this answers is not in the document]
alee: Yep,  probably good to clean that up - somehow there are 337 members in the Reviewers group on prod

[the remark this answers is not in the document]
ldwiyono: I put a single (non-static) method `is_group_member` which can be called on all instances of ComsesGroups (e.g. `ComsesGroups.ADMIN.is_group_member(user)`. I'm not sure if this clashes with Karthik's work on the PeerReviewer object.

ldwiyono:
I think anyone with moderator perms should now be able to access the "Mark spam" button. The `is_group_member` method of ComsesGroups that I added is not static and may need adjustment.

django/core/models.py
```diff
@@ -57,7 +57,14 @@ def get_group(self):
         if _group is None:
             _group = self.group = Group.objects.get(name=self.value)
         return _group
-
+    
+    # can be called like #ComsesGroups.ADMIN.is_group_member(user)
+    def is_group_member(self, user):
+        for member in self.users():
+            if member.username == user.username:
+                return True
```
sgfost, line 65: IMO the instance method approach (as opposed to multiple static methods) is fine, though this implementation is less efficient than the direct query (`user.groups.filter(name=self.value).exists()`) that its meant to act as a convenience method for

alee: Yep, also @ldwiyono make sure your git commits follow the conventional commits guidelines:

https://www.conventionalcommits.org/en/v1.0.0/#summary

ldwiyono: Yes, sorry, that is inefficient. I will try and push the more efficient `return self.users().filter(username=user.username).exists()`.

alee: I stuck with the static methods approach for now, just pushed a commit with some adjustments to the boolean logic in view_helpers.py as well

alee (approved):
LGTM, thanks @ldwiyono and @sgfost !

PR #757, opened 2024-08-27
add note about frozen files on peer review completion and remove duplicated review reminders content

sgfost:
PR'd to confirm the wording is ok: https://github.com/sgfost/comses.net/blob/8dfbd55f099548114ba1f12da008fcbd226e33e6/frontend/src/components/ReviewReminders.vue#L15-L18


frontend/src/components/ReviewReminders.vue
```diff
@@ -0,0 +1,69 @@
...
+    <p>
+      <i class="fas fa-exclamation-triangle"></i> Peer reviews must be requested on
+      <strong>unpublished model releases</strong> so that you can continue to revise your submission
+      based on reviewer guidance.
+    </p>
+    <p>
+      If you have already published your model
+      <strong>a new draft release will be created</strong> as an exact copy of your most recent
+      published release. If your model successfully passes peer review you can then choose to
+      publish the reviewed release as the latest version.
```
alee, line 12: Should we add a sentence that emphasizes that they will be able to change the version as they publish their model?

alee (approved):
LGTM, glad to have the refactor to remove the duplicated text as well!

frontend/src/components/ReviewReminders.vue
sgfost: Yeah good idea, will do. I'm also going to see if I can do anything to reduce the whole thing's wall-of-text-ness while I'm at it

sgfost:
looks like this now

image

still lots to read through but I think its fairly close to as slim as it can be

alee (approved):
all of this LGTM but I'd like to take the time to fix the codemeta hypothesis tests here or die trying so holding off on the merge / deploy until I get to the bottom of some things in test_codemeta.py

sgfost:
I noticed 2 things

- in `CodeMetaValidationTest` the contributors are popped from the `users` set but the teardown method is trying to delete contributor users by referencing that set that no longer has those users in it
- there doesn't seem to be a uniqueness constraint on the usernames drawn for the rest of the tests in `CodeMetaTest` other than `test_authors`. I didn't see any logs from a sample of test failures that indicate the error was coming from there, though

alee:
> I noticed 2 things
> 
> * in `CodeMetaValidationTest` the contributors are popped from the `users` set but the teardown method is trying to delete contributor users by referencing that set that no longer has those users in it
> * there doesn't seem to be a uniqueness constraint on the usernames drawn for the rest of the tests in `CodeMetaTest` other than `test_authors`. I didn't see any logs from a sample of test failures that indicate the error was coming from there though

Just chatted with @asuworks and we're going to rip out all of the CodeMeta stuff from main and put it in properly in #699 - thanks for taking a look Scott!

PR #766, opened 2024-09-26
fix: set initial ordering to `relevance` in the frontend component

asuworks:


frontend/src/components/CodebaseListSidebar.vue
```diff
@@ -141,7 +141,7 @@ const initializeFilterValues = () => {
   values.endDate = urlParams.get("publishedBefore")
     ? new Date(urlParams.get("publishedBefore")!)
     : null;
-  values.ordering = urlParams.get("ordering") || "-first_published_at";
+  values.ordering = urlParams.get("ordering") || "relevance";
```
alee, line 144: Can you explain what behavior this is attempting to fix? The initial sort order should be newest to oldest as "relevance" doesn't make sense on initial load; won't this set the initial sort order to "relevance" when the ordering query param isn't explicitly set?

asuworks: If the user sets the ordering explicitely, the ordering will be available through `urlParams.get("ordering")`. If not, the default ordering during searching (seqrch query is not empty) is "relevance".

But I guess, you also have a point, namely: when there is no search query, just filtering by programming languages. In which case it should be something like:

```
values.ordering = urlParams.get("ordering") || (urlParams.get("query") ? "relevance" : "-first_published_at");
```

alee (approved):
I didn't see an easy way to hide "Relevance" based on the way ListSortBy.vue and codebase_list.ts are currently set up but if you can think of that would be ideal

asuworks:
https://github.com/comses/comses.net/commit/4ed8cd87fc4ebb7b25f2e5cb5b46ab02c7f3f729

alee:
looks good to me, go ahead and submit a new PR for it and we'll get it merged and deployed

PR #769, opened 2024-10-22
fix: remove `Relevance` when `query` parameter is empty for `/codebases/` search

asuworks:


sgfost:
this branch is a little borked because of the last pr getting squashed and then reverting the requirements.txt change. Can you just cherry-pick fdfacaa?

asuworks:
Thanks, Scott!

sgfost:
sure, I believe Allen had a comment about the spread thing on the old commit. I don't think its super unreadable personally but I guess

```js
let opts = [2,3,4];
if (hasQuery()) opts = [1, ...opts] ;
```

is slightly more readable

asuworks:
> sure, I believe Allen had a comment about the spread thing on the old commit. I don't think its super unreadable personally but I guess
> 
> ```js
> let opts = [2,3,4];
> if (hasQuery()) opts = [1, ...opts] ;
> ```
> 
> is slightly more readable

oh, did I kill it?? 

You mean here?
```  ...(hasQueryParam("query") ? [{ value: "relevance", label: "Relevance" }] : []),
```

sgfost:
yes

django/library/views.py
```diff
@@ -399,12 +399,13 @@ def filter_queryset(self, request, queryset, view):
             criteria.update(id__in=codebases.values_list("id", flat=True))
             # or we could include the PL in the query
             # qs += " ".join(programming_languages)
+
+        default_ordering = "-first_published_at"
         if ordering:
```
alee, line 404: I think this conditional could probably be simplified also, is this equivalent?

```
        default_ordering = "relevance" if qs else "-first_published_at"
        criteria.update(ordering=ordering if ordering else default_ordering)
```

alee (approved):
LGTM, it's fine to leave the spread operator if you all prefer, it is a bit more concise

django/library/views.py
asuworks: I asked Perplexity if the approaches are equivalent. Crazy, it does execution! I have also checked the answer. It is correct.

# Perplexity Response
The two approaches are not equivalent, as demonstrated by the execution results of the provided code snippets.
## Approach 1:
```python
default_ordering = "-first_published_at"
if ordering:
    criteria.update(ordering=ordering if qs else default_ordering)
else:
    # set default ordering for search when ordering is not specified
    criteria.update(ordering="relevance" if qs else default_ordering)
```
### Execution Results:
```
(True, None, 'relevance')
(True, 'some_order', 'some_order')
(False, None, '-first_published_at')
(False, 'some_order', '-first_published_at')
```
## Approach 2:
```python
default_ordering = "relevance" if qs else "-first_published_at"
criteria.update(ordering=ordering if ordering else default_ordering)
```
### Execution Results:
```
(True, None, 'relevance')
(True, 'some_order', 'some_order')
(False, None, '-first_published_at')
(False, 'some_order', 'some_order')
```
## Key Differences:
When qs is False and ordering is 'some_order':
Approach 1 results in `-first_published_at`.
Approach 2 results in `some_order`.

This difference arises because Approach 1 checks whether ordering should be used based on the value of qs, while Approach 2 uses ordering directly if it is provided. Therefore, the logic for determining the final ordering differs between the two approaches when qs is False and an ordering is specified.

asuworks: TL;DR if no `query` is given, the ordering should be `-first_published_at` and not `some_order` which might be present in request params

alee: That's cute that the AI can generate truth tables! It doesn't actually answer the question of "is it the behavior we want" though.

Four cases:

1. `ordering` is set, `qs` is not - should be set to `ordering` right? not the default ordering
2. `ordering` is not set, but `qs` is - set to "relevance"
3. `ordering` is set, `qs` is set - respect ordering requested by the user again
4. `ordering` not set, `qs` is not set - use "-first_published_at"

Here's the current truth table for approach 1:

| ordering | qs | criteria['ordering'] |
| ------ | ------- | ------ |
| True | True | ordering |
| True | False	| default_ordering |
| False	| True | 	"relevance" |
| False | 	False	| default_ordering |

Shouldn't the second line be set to `ordering` even though `qs` is empty?

alee:
closing in favor of up-to-date branch #770

django/library/views.py
asuworks: agree!

PR #770, opened 2024-10-24
remove relevance in sort order when no query string is present

alee:
cherry-picked anton's commit 376bc2715b9e8d2e22e2c4e518faee4c587c6d16 from #769 and laid over main instead of the defunct squashed + merged branch


alee (commented):
thanks Anton! Is the entire `clearAllFilters` machinery used? I don't think it gets called by anything live in ListSidebar.vue or elsewhere.

Setting isFilterChanged to true appears to enable Apply Filters always instead of only when there are filters to apply though...

PR #772, opened 2024-11-16
add spam detection endpoints for external LLM service

asuworks:
Add spam detection endpoints: `api/spam/get-latest-batch/` and `api/spam/update`. 

LLM Spam Detection Process:
1. A SpamModeration record with status `SCHEDULED_FOR_CHECK` is stored on every Job, Event, Codebase submission.
2. A decoupled external service [asuworks/comses.spamcheck](https://github.com/asuworks/comses.spamcheck) queries for these SpamModeration objects, analyzes them for spam and submits a spam report to `api/spam/update` for each one of them.
3. The handler for the `api/spam/update` on the CoMSES side updates the corresponding SpamModeration object according to the LLM report from the external service.


sgfost:
just missing the migration for `SpamModeration.status`, which I think should turn anything with "unreviewed" status into "spam_likely" status.

I'll read through it again but it seemed all good on a first pass, besides having a way to kick off the process

asuworks:
Regarding a way to start the process from CoMSES side: 
1. start the instance with openstack-cli
2. poll the service until `HEALTHY`
3. keep polling the service untill the workflow is `COMPLETED`
4. shutdown instance with openstack-cli

something like this?

sgfost:
@asuworks I just remembered there was some additional cleanup I wanted to do eventually with the spam stuff. This might be a good place to get that done if you are up for it. comses/planning#249. Namely the second point (refactoring the serializer mixin to actually be just a mixin)

django/curator/management/commands/curator_llm_spam_check.py
sgfost, line 1: can this (everything besides `Command`) be moved to a module in curator/? Maybe `spam.py`

asuworks:
added "one-click" install of the `comses.spamcheck` with ansible: to https://github.com/asuworks/comses.spamcheck/deploy

The script does the following:
1. basic server configuration: harden ssh, fail2ban
2. install ollama and load llama3.1, llama3.2
3. clone and run `comses.temporal`
4. clone and run `comses.spamcheck`

After the ansible playbook is done, the management commands from CoMSES should be able to trigger `CheckSpamWorkflow`.

PR #773, opened 2024-11-26
add user map of institutions (WIP)

CharlesSheelam:
-seperated codebase and codebase release total counts on the metrics table
-set up the backend to gather institution data for the user map


sgfost (changes requested):
mostly structural stuff. LMK when the map is working and I'll take another look

sgfost (commented):
We should capture the additional metadata from RoR when we make new entries. This should be an easy fix and I can do it in the next few hrs along with (hopefully) finding a better layout for the metrics page 

Everything else looks good. Thanks @CharlesSheelam and @alee for fixing.

sgfost:
looks like this now:

Screenshot from 2024-12-11 20-11-56

alee (approved):
thanks @sgfost & @CharlesSheelam ! deployed on staging and lgtm

PR #778, opened 2024-12-05
deps: npm update

alee:
also resolves https://github.com/comses/planning/issues/287


sgfost (approved):
LGTM, those were the only two mentions I found as well

PR #784, opened 2024-12-20
fix: order MemberProfile search results by relevance when searching `/users/?query=...`

asuworks:


sgfost (commented):
well thats an easy fix. Should `order_by_relevance` just default to `True`? Categorically, there are only two cases where we don't want to order by relevance (no query at all OR explicit "sort by"), right?

asuworks:
yeah :) yes, it could be set to True by default

alee:
hey @sgfost - any ideas why the e2e build is failing? I got as far as viewing the cypress videos but would need more focused time to look at how the tests are testing to see why it's failing to find the expected text

alee:
It looks like the sort by ordering got messed up for Events (probably Jobs too), I'll look into it.

PR #787, opened 2025-01-06
Password reset link on member profile page + winter '24 digest update

sgfost:
button is added to the profile sidebar:

image


alee (approved):
thanks Scott! I think we should move to just linking to the OSF digest component like you suggest in https://github.com/comses/planning/issues/303

PR #790, opened 2025-01-09
refactor metadata generation and add huey for async task processing

sgfost:
#### additions

- adds `huey` for async task processing
- replaces codemeta generation with `library.metadata.CodeMetaConverter` which uses `codemeticulous` and comes with some updates to the transformation
- adds citation file format generation with `library.metadata.CitationFileFormatConverter`
- adds `library.metadata.DataCiteConverter`
  - intended to replace `DataCiteSchema`, but does not yet do so. Should happen with comses/planning#286, *this can either happen in this PR or later*
- `codemeta_snapshot` is used as the primary accessor for codemeta (and anything derived from it)
  - a json field on `Codebase` and `CodebaseRelease` that is updated when the object is saved
  - keeps a 'cached' version for quicker access (mostly relevant for the release detail page)
  - allows for an easy way to track when changes happen to metadata in order to trigger events
- whenever a change in codemeta is detected for a release, the release package will be rebuilt with new metadata files
- `LICENSE` and `CITATION.cff` files are now included in release packages

#### deploying

since codemeta is no longer re-generated each time it's accessed, it needs to be seeded with `./manage.py update_codebase_metadata`, which will build and store a codemeta representation for every codebase/release. This shouldn't ever need to be ran again unless there are changes to the codemeta generation in the future that we want to be applied everywhere


alee:
No need for major changes but did ya see the latest wagtail release with integrated support for Django-tasks?

sgfost:
Oh interesting, I looked at that back when trying to figure out what to use. Still in alpha (only comes with a db backend and missing a couple things like retries)

The DB/ORM backend might have potential to be more broadly useful than huey + redis so is something to consider. Right now I'm only really using it for things that already have some sort of result storage already baked in (e.g. result of a `push_to_github` task is reflected in `GitRemote` model) so the quickness + ephemeral-ness of redis won out

Seems like it'll make it into django 6.0 and the api is very similar to huey

django/library/tests/test_views.py
```diff
@@ -562,7 +562,7 @@ def test_publish_codebaserelease(self):
         self.assertRaises(ValidationError, lambda: self.codebase_release.publish())
 
         self.codebase_release.programming_languages.add("Java")
-
+        self.codebase_release.save()
```
alee, line 565: is this needed before `publish()` which also saves?

django/library/views.py
```diff
@@ -822,6 +822,9 @@ def contributors(self, request, **kwargs):
         )
         crs.is_valid(raise_exception=True)
         crs.save()
+        # re-generate codemeta
+        codebase_release.codebase.save(rebuild_metadata=False)
```
alee, line 826: should this be `rebuild_metadata=True` based on the preceding comment?

alee (commented):
looks like a solid improvement, some minor comments included inline

great work as usual @sgfost !

django/library/metadata.py
```diff
@@ -0,0 +1,387 @@
...
+            url=affiliation.get("url"),
+            identifier=affiliation.get("ror_id") or None,
+        )
+
+    @classmethod
+    def _convert_actor(
+        cls, contributor, actor_type: Literal["author", "contributor"], index: int
+    ) -> Person | Organization:
+        # https://www.w3.org/TR/json-ld11/#identifying-blank-nodes
+        actor_id = contributor.orcid_url or f"_:{actor_type}_{index + 1}"
```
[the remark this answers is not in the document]
sgfost: definitely, thanks. somehow missed that

django/library/views.py
sgfost: yep, I believe I meant to say `rebuild_release_metadata=False`

PR #791, opened 2025-01-09
git mirroring

sgfost:
splitting #718


sgfost:
I believe all of the main functionality for the 2-way sync is now implemented and works for simple cases. Likely not bulletproof nor polished so I now need to do lots and lots of testing

PR #793, opened 2025-01-23
deps: replace jwt with pyjwt

sgfost:
This is the correct one but doesn't actually make any difference currently, since the `jwt` module only gets loaded by the allauth google provider, [which is installed but not used](https://github.com/comses/comses.net/blob/67cda2fb78520a004f9105f72d42f132d5813c50/django/core/settings/defaults.py#L130)


alee:
did things work if jwt was removed?

PR #795, opened 2025-03-02
feat: new landing page layout

CharlesSheelam:
-new landing page layout based off of the latest draft in: https://github.com/comses/planning/issues/275
-placeholders in place for events, jobs, forum and youtube video feeds from @Karthik99999

### DIFF
--- a/django/home/jinja2/home/index.jinja
+++ b/django/home/jinja2/home/index.jinja
@@ -5,22 +5,20 @@
         <div class='container'>
             <div class='row'>
                 <div class='col'>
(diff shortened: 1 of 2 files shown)


CharlesSheelam:
looks like this now:

image
image
image
image

CharlesSheelam:
landing page  looks like this now. only thing left if everything else looks good is to finalize a background pattern design so that its not repeating like in the picture

image
image
image
image

CharlesSheelam:
looks like this now:

-updated the background image so that it doesn't repeat akwardly
-fixed the margins of the logo on the top left and other headers

image

PR #812, opened 2025-05-19
fix datacite metadata conversion failing and citation authors out of order

sgfost:
the datacite fix is straightforward, just added the property back

the problem with the authors being out of order is that we were 'converting' from a `ReleaseContributor` queryset to a `Contributor` queryset via `id__in=rc_qs.values("contributor_id")` which lost the ordering. This helper with Case/When preserves that order and seems to be a common-enough idiom https://stackoverflow.com/a/37648265

### DIFF
--- a/django/library/models.py
+++ b/django/library/models.py
@@ -21,7 +21,7 @@ from django.core.cache import cache
 from django.core.files.images import ImageFile
 from django.core.files.storage import FileSystemStorage
 from django.db import models, transaction
-from django.db.models import Prefetch, Q, Count, Max
+from django.db.models import Prefetch, Q, Count, Max, Case, When
 from django.utils.functional import cached_property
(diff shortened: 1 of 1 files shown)


alee (commented):
Thanks Scott! I went for a different approach in #813 where we don't care about ordering when asking a Codebase for all of its contributors (set union) and make sure to enforce it in a CodebaseRelease. 

If we decide to implement ordering in the Codebase citable authors we might use something like your _get_ordered_contributors but I feel like whatever ordering we impose will still have corner cases so inclined to leave it as unordered for the time being

sgfost:
oh nice, that is way more intuitive, didn't occur to me. The only reason I can think of wanting ordered codebase contributors is for codemeta/datacite representations but I think these are mostly internal

PR #815, opened 2025-05-24
GitHub integration

sgfost:
adds a feature to the CML which allows for:
- **connecting** a public GitHub repository with a codebase in the CML and;
- **pushing** pushing a git repository constructed from public CML releases to the connected GitHub remote
- **importing** individual releases made on a GitHub repository into the CML as new releases (of a special "imported" type) which can be published/peer reviewed

Setup/management is done by the submitter on a configuration page for each codebase, accessible from the sidebar of a codebase page or the feature overview page (`/github/`). From there, one is walked through the process including connecting a github account with the existing oauth flow and installing our github app on the same account, then given the option of:
- connecting with a **new repository**: allows connecting to a blank github repository, then generates a git repo for a codebase which can be pushed (codebase updates are reflected in the git repo and the push can be repeated). Importing GitHub releases back is also allowed
- connecting with an **existing repository**: allows connecting with an existing GitHub repo and importing GitHub releases as codebase releases

### configuration

-> https://github.com/comses/infra/wiki/GitHub-Sync

#### new models

- `GithubIntegrationAppInstallation`: tracks the github app installation status for a user
- `CodebaseGitRemote`: represents a remote repository that should be pushed to and/or checked for new releases
- `GitRefSyncState`: represents the state of a git ref (commit/tag/branch) that was generated for a codebase and the state of the push to GitHub
- `ImportedReleaseSyncState`: represents the state of a release imported from GitHub, the existence of this means a release was imported from github, and should be treated differently in some cases
- `GitHubIntegrationConfiguration`: (admin dashboard-editable) feature settings including overview page content

#### github APIs

- `GitHubApi`: main wrapper for github operations (auth, pushing, creating releases, etc.)
- `GitHubReleaseImporter`: handles importing releases from github by downloading the archive, extracting metadata, and creating a new CodebaseRelease

#### filesystem

- `CodebaseGitRepositoryApi`: manages local git repository operations like building and updating a git repo from a codebase's (public) releases
- `FileCategories`/`CategoryManifestManager`: file categorization changed in order to handle imported releases which are not organized into category directories, information is instead kept in a manifest in the db and updated by the submitter
- `ImportedCodebaseReleaseFsApi`: sibling filesystem API of the regular `CodebaseReleaseFsApi` with some special handling and additional functionality
- adds generated `LICENSE` and `CITATION.cff` files to new release packages

#### webhooks

`library.views.github_sync_webhook`: the github app sends webhook events to the server and we respond to installation events to track user installations


deploy/conf/.env.template
```diff
@@ -45,6 +45,12 @@ DATACITE_DRY_RUN="true" # allowed values: "true" or "false"
 # youtube api settings
 YOUTUBE_CHANNEL_ID=
 
+# github integration app
+GITHUB_INTEGRATION_APP_ID=
+GITHUB_INTEGRATION_APP_NAME=
+GITHUB_INTEGRATION_APP_INSTALLATION_ID=
```
alee, line 51: should document INSTALLATION_ID on the infra wiki: purpose, how / when is it set in dev / staging / prod, etc

django/core/tests/base.py
```diff
@@ -187,5 +187,13 @@ def initialize_test_shared_folders():
     )
 
 
+def clear_test_shared_folder(dir=settings.REPOSITORY_ROOT):
+    for fs in os.scandir(dir):
+        if fs.is_dir():
+            shutil.rmtree(os.path.join(dir, fs.name), ignore_errors=True)
+        elif fs.is_file():
+            os.remove(os.path.join(dir, fs.name))
```
alee: should standardize new code to favor pathlib

django/library/models.py
```diff
@@ -3347,3 +3883,166 @@ def __str__(self):
                    HTTP Status: {self.http_status}, Message: {self.message},
                    Hash: {self.metadata_hash}, DOI: {self.doi}
                    """
+
+
+class GitHubIntegrationFaqEntry(index.Indexed, Orderable):
```
alee, line 3894: Minor nitpick that is a non-blocker, we should be consistent about naming (e.g., `GithubIntegrationAppInstallation`). I think I tend to favor `Github` for Python symbols to NotCapitalizeAllTheThings and GitHub in docs etc. but since there are many more of these it's probably easier to just rename `GithubIntegrationAppInstallation` and then redo the schema migrations.

frontend/src/components/githubIntegration/AccordionStep.vue
alee, line 1: still not a fan of camel casing in the directory (`releaseEditor` suffers from the same issue), would prefer `release-editor` and `github-integration` on the filesystem

frontend/src/components/githubIntegration/ReleaseManagementSection.vue
```diff
@@ -0,0 +1,406 @@
...
+  const ghDate = gh?.publishedAt || gh?.createdAt;
+  const localDate = (local as any)?.dateCreated;
+  const raw = ghDate || localDate;
+  if (!raw) return 0;
+  const d = raw instanceof Date ? raw : new Date(raw);
+  const t = d.getTime();
+  return Number.isNaN(t) ? 0 : t;
+}
+
+function sortEntries(matrix: ReleaseAlignment) {
```
alee, line 264: might be worth bringing in the npm `semver` or `semver-ts` library for this instead if we want to deal with semver corner cases but this is probably fine for initial release

alee (approved):

PR #820, opened 2025-06-20
refactor: Index digest records from Zenodo community

Karthik99999:
Changes https://comses.net/digest to index from https://zenodo.org/communities/comses-digest/records rather than locally hosted files. Still need to handle pagination in the api so drafting for now. I also mostly reused the logic for adding the `ComsesDigest` objects by depending on the file name of the Zenodo upload, but this could probably be changed to use some metadata on Zenodo instead.


django/home/management/commands/update_digest_archive.py
```diff
@@ -6,45 +6,43 @@
...
-            logger.error(err_msg)
-        else:
-            logger.info(f"\n\nSuccessfully indexed all pdfs in {DIGEST_STATIC_DIR}")
+        logger.info(f"\n\nSuccessfully indexed all pdfs from Zenodo community.")
 
-    def add_digest_archive(self, file_name):
+    def add_digest_archive(self, file_name, url):
         # FIXME: consider rolling all these regex searches into a single-pass regex monstrosity
         volume = int(re.search(r"vol(\d+)", file_name, re.IGNORECASE).group(1))
         issue_number = int(re.search(r"no(\d+)", file_name, re.IGNORECASE).group(1))
```
sgfost, line 48: It'd be nice to not rely on the filename from zenodo to pull out the volume/issue number/season. There isn't really a good way to put this in the zenodo metadata but one option would be just putting that in the description and displaying whatever the zenodo description is instead of formatting the model fields

sgfost (approved):

alee (approved):
looks great, thanks @Karthik99999 and @sgfost !

PR #821, opened 2025-07-10
feat: use huey to schedule doi minting

alee:
use huey to schedule DOI minting asynchronously after publishing a peer reviewed model that doesn't have a DOI

there's some badness with circular imports in library/doi.py and library/models.py as a result of this but didn't see a quick way around that


sgfost (commented):
unfortunately there's no good way around the function-level imports for tasks without doing liposuction on the models

alee:
build currently failing on vite not completing successfully due to

```
vite-1  |
vite-1  | > frontend@1.0.0 build
vite-1  | > run-p type-check build-only
vite-1  |
vite-1  |
vite-1  | > frontend@1.0.0 type-check
vite-1  | > vue-tsc --noEmit -p tsconfig.vitest.json --composite false
vite-1  |
vite-1  |
vite-1  | > frontend@1.0.0 build-only
vite-1  | > vite build
vite-1  |
vite-1  | failed to load config from /code/vite.config.ts
vite-1  | error during build:
vite-1  | Error: ENOENT: no such file or directory, scandir '/code/src/apps'
vite-1  |     at Object.readdirSync (node:fs:1583:26)
vite-1  |     at getAppEntries (/code/vite.config.ts:46:21)
vite-1  |     at Object.<anonymous> (/code/vite.config.ts:79:14)
vite-1  |     at Module._compile (node:internal/modules/cjs/loader:1688:14)
vite-1  |     at _require.extensions.<computed> [as .js] (file:///code/node_modules/vite/dist/node/chunks/dep-827b23df.js:66675:24)
vite-1  |     at Module.load (node:internal/modules/cjs/loader:1423:32)
vite-1  |     at Function._load (node:internal/modules/cjs/loader:1246:12)
vite-1  |     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
vite-1  |     at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
vite-1  |     at Module.require (node:internal/modules/cjs/loader:1445:12)
vite-1  | ERROR: "build-only" exited with 1.
```

sgfost:
> e2e build currently failing on vite not completing successfully

the resulting docker-compose.yml from `make docker-compose.yml` now that `DEPLOY_ENVIRONMENT=test` for the e2e build doesn't contain a volume bind mount for the contents of `frontend/` or `frontend/src/`

though I'm not sure where in the hierarchy it makes sense to add this (base.yml/test.yml/e2e.yml)

sgfost:
got it, the e2e server was trying to run `test.sh` instead of dev.sh (manage.py runserver) now that it extends `test.yml`

I'll push the minimal fix in a moment to make sure it works here too

sgfost:
that may be the two migrate processes colliding, thats what I saw locally anyways, which prompted removing one but then inv prepare was running before migrating finished. If so that leaves either adding a `wait 42` like make deploy, e2e.sh that doesn't db.init, or making the healthcheck wait on migrate somehow

alee:
> that may be the two migrate processes colliding, thats what I saw locally anyways, which prompted removing one but then inv prepare was running before migrating finished. If so that leaves either adding a `wait 42` like make deploy, e2e.sh that doesn't db.init, or making the healthcheck wait on migrate somehow

Ah that's right, since it's running dev.sh now it should automatically migrate. A `sleep 42` would work but I wonder why this issue is popping up now though?

sgfost:
this works: https://github.com/sgfost/comses.net/commit/2fb0b2c843806d92202f45d739f4d9c594538a30

PR #834, opened 2025-11-07
update peer review passed email

alee:
add call to contact us if interested in participating in the ten-minute dialog video series and minor text refinement

closes https://github.com/comses/planning/issues/341

### DIFF
--- a/django/library/jinja2/library/review/email/model_certified.jinja
+++ b/django/library/jinja2/library/review/email/model_certified.jinja
@@ -5,14 +5,15 @@
 
 Dear {{ review.submitter.name }},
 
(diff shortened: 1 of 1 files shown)


seanbergin (approved):
That's great. Will this be up immediately? Will sending things to the contact us email be annoying in the long term since the emails will be forwarded to someone else? Or is this so low volume that its not going to be an issue?

alee:
> That's great. Will this be up immediately? Will sending things to the contact us email be annoying in the long term since the emails will be forwarded to someone else? Or is this so low volume that its not going to be an issue?

Will be up as soon as we merge and deploy the next release - as for spamming the contact form, we could set up a dedicated email that this goes to instead or a more custom workflow to manage the request queue but this can always be adjusted if / when it becomes a problem.

PR #869, opened 2026-04-07
chore: add AGENTS.md

sgfost:


alee (commented):
Thanks @sgfost ! I asked copilot, claude, and chatgpt to review this file and ask about general best practices for AGENTS.md + recommendations to avoid drift and they seem to have mostly converged on making the AGENTS.md more policy focused and non-obvious invariant focused instead of having a set of commands. I'm leaning towards following that advice but wanted to see what you and @asuworks think about this.

You can see the full diff at https://github.com/alee/comses.net/commit/1938cb88fe12a1e41000df37512a8d0b5481900f

alee (approved):

asuworks:
Is there any particular reason not to make RUNBOOK, PLAN, HANDOFF and CHECKPOINT skills?
docs don't seem to be the right place to store this.

Content-wise seems ok.

alee:
that's a good point about agent skills, haven't really touched on that yet. is there an agent-agnostic way to "store" skills in a repository? 

`docs/agents` is a place for durable agent things, so a skills folder there might work. then agents.md points at the skills folder for discoverability?

here's what codex had to say about building a portable skill contract that could be persisted in a repo.

```
Short answer: not instead, but possibly in addition.

For what you have now, keeping these as repository docs is the right default:

CHECKPOINT.md
HANDOFF.md
PLAN.md
RUNBOOK.md
README.md
Why:

These are durable project artifacts used by humans and agents.
They are simple, reviewable, and easy to version in normal docs flow.
They do not require agent runtime behavior to stay useful.
When a skill makes sense:

You want an executable workflow (for example, “create new workstream folder and prefill handoff/checkpoint files”).
You want repeatable automation with trigger phrases and guidance logic.
You want slash-command style task orchestration, not just storage templates.
Can skills be persisted in a repository?
Yes. Repository-scoped skills are persisted by committing them into the repo, typically under one of these locations:

.github/skills/<skill-name>/SKILL.md
.agents/skills/<skill-name>/SKILL.md
.claude/skills/<skill-name>/SKILL.md
Recommended approach for this repo:

Keep docs templates as the canonical source in templates.
Optionally add one small skill that scaffolds a new workstream from those templates.
Keep policy in AGENTS.md and operational artifacts in agents.
```

The skill templates it generated:

```
# Portable Skill Contract (Agent-Agnostic)

Use this as the canonical definition of a reusable skill workflow. Keep this file vendor-neutral and map it into tool-specific adapter files.

## Skill Identity

- Skill name:
- Version:
- Owner:
- Last updated (UTC):

## Intent

- Problem this skill solves:
- Trigger phrases users might use:
- Inputs expected:
- Outputs produced:

## Preconditions

- Required files/folders:
- Required tools/commands:
- Required permissions:

## Workflow

1. Step 1:
2. Step 2:
3. Step 3:

## Safety Constraints

- Security constraints:
- Data integrity constraints:
- Permission constraints:
- Non-destructive defaults:

## Validation

- Checks to run:
- Success criteria:
- Failure handling:

## Artifacts

- Files created/updated:
- Logging/checkpoint expectations:
- Handoff expectations:

## Non-Goals

- Explicitly out of scope behavior:

## Adapter Notes

- Keep this file as source of truth.
- Tool-specific adapters should only add discovery metadata and map to this workflow.
- Avoid duplicating business logic across adapters.
```

asuworks:
There are essentially 3 objects we're talking about when describing a multi-session collaboration with agents: 
1. skills
2. cli/mcp tools
3. artifacts

**skills** tell the agent how to use **tools** (optional) to build **artifacts** and where to store them.

Idea: we could keep some tooling (linting/formatting/etc...) and path conventions in docs/ but move your templates into the skills files. Conventions in docs/ = guardrails = instructions for agents about where to store: handoffs, plans, ADRs, etc.. some devs might use their own skills/harnesses, so we want to define what should be stored where. These conventions must be referenced in the root AGENTS.md

Or am I butchering the intent of your templates with this?

---

We should use https://agentskills.io/skill-creation/quickstart so developers can install and use these skills with their preferred agent using `npx skills add comses/comses.net` (if we keep skills in comses/comses.net/skills - which is not a bad idea i think). This is what I used for [lazyodd.](https://github.com/asuworks/lazyodd/)

PR #874, opened 2026-05-26
feat: Implement youtube, citation, and input/output data schema additions

Karthik99999:
comses/planning#332

Currently does not include changes to the Upload Form Page to include fields to edit input/output urls, still figuring out issues with actually saving the data on that page.


alee:
added in #890 and #891

PR #881, opened 2026-07-28
Add CoMSES Librarian single sign-on and member access points

asuworks:
Lets CoMSES members sign in to the Librarian with their comses.net accounts, gated on a new `Librarian Users` group, and surfaces it on the homepage and in the Model Library menu for members. Mirrors `discourse_sso`.

**Off until configured.** With `LIBRARIAN_BASE_URL` unset — the default, and what `.env.template` ships — `/librarian/sso` returns 404 and the menu entries stay hidden.

**After merge:** set `LIBRARIAN_BASE_URL` per deployment, and generate a `librarian_sso_secret` per deployment, ≥32 random bytes. Separate values matter: a shared one would let the staging secret mint production sessions.


asuworks:
both suggestions applied

alee (approved):
looks solid, thanks Anton!

I fixed the discourse staging server and discourse_sso still appears to work after the last batch of commits. If the librarian_sso endpoint still works for your things, I'm ready to merge this and we can move on to the next set @asuworks

asuworks:
librarian sso still works on staging, thanks @alee!

PR #884, opened 2026-07-31
feat: Allow YouTube urls in media section of models

Karthik99999:
Separating from #874

Also adds emitted events to the `BootstrapModal` class to make it simpler to pause the video when the media modal is closed.


alee (approved):
thanks @Karthik99999 !

PR #890, opened 2026-08-14
feat: Add Input Data URL field in release metadata

Karthik99999:
#874 

Did not add the text fields to the upload data section, as doing so requireed making a lot of annoying edits/hardcodes to UploadFormPage. I can still make the changes if the ability to update input/output data urls needs to be available in that section of drafting a new release.


alee (approved):
easy to review, thanks @Karthik99999

let me know when you're ready for the next set of things

